PR #7, opened 2023-06-30
Add basic unit testing for pam and nss against the himmelblau daemon

dmulder:


dmulder:
These tests expose a bug in nss, but I'll merge this anyway pending a fix for the nss issue.

PR #46, opened 2023-09-25
Enrollment: ConfidentialClient via device add and application certificate

dmulder:
This provides simple device enrollment and confidential authentication. This loosely follows [MS-DVRJ], [MS-DVRD], and [MS-OAPXBC]. Where I diverge from the spec is where Microsoft does also. I've filed cases with doc-help requesting fixes for the documentation which is outdated.

Update: Microsoft has rejected the doc updates, and refuses to document Azure cloud related protocol changes.

Checklist

- [x] This pr contains no AI generated code
- [x] cargo fmt has been run
- [x] cargo clippy has been run
- [x] A functionality test has been added
- [x] make test has been run and passes


src/msal/src/misc.rs
```diff
@@ -149,8 +232,141 @@ pub async fn enroll_device(graph_url: &str, access_token: &str) -> Result<Device
...
+    };
+    let url = &format!("{}/v1.0/applications/{}", graph_url, app_id);
+    let mut key_creds = match app.key_creds {
+        Some(key_creds) => key_creds,
+        None => Vec::<KeyCredential>::new(),
+    };
+    key_creds.push(KeyCredential {
+        custom_key_identifier: None, /* Leaving this blank will default to the thumbprint */
+        display_name: Some(desc.to_string()),
+        end_date_time: None, /* Leaving this blank defaults to 1 year from now */
```
dmulder: This is the most serious concern. MS docs say that this certificate will expire after a single year. The Azure UI actually interprets this as expiring in the year 4095, so I don't know which is correct.

src/common/src/cert.rs
```diff
@@ -0,0 +1,39 @@
+use crate::constants::DEFAULT_CERT_DIR;
+use anyhow::Result;
+use std::fs;
+use std::fs::File;
+use std::io::Write;
+use std::path::Path;
+
+/* TODO: The private key at least needs to be stored in the cache and encrypted */
```
dmulder: This needs work. We should store the device private key somewhere more secure.

dmulder: Will need to work with Kanidm to add these to the storage. This MR is **blocked** until we have a proper way of storing the certificate and private key. This WIP request is storing the private key readable by anyone. Obviously this is wrong. Will need to send the key to the daemon and have it stored securely, with only the daemon having read permissions.

dmulder: Need to use [kanidm/hsm-crypto](https://github.com/kanidm/hsm-crypto).

dmulder:
I'm currently performing some tests on this branch and fixes bugs. Otherwise this is mostly finished.

dmulder:
This MR is now only waiting on 3 external issues (and possibly additional testing/modifications when these are fixed):

1. Himmelblau requires access to the keystore at auth time (kanidm [MR#2385](https://github.com/kanidm/kanidm/pull/2385)).
2. Kanidm's `db.insert_tagged_hsm_key` isn't actually caching the hsm key (kanidm [Bug#2387](https://github.com/kanidm/kanidm/issues/2387)).
3. msal `remove_account()` function fails with "TypeError: string indices must be integers, not 'str'". This prevents himmelblau from removing a cached account when the PRT request fails (msal [Bug#643](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/643)).

dmulder:
> * [x]  msal `remove_account()` function fails with "TypeError: string indices must be integers, not 'str'". This prevents himmelblau from removing a cached account when the PRT request fails (msal [Bug#643](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/643)).

This one was actually a bug in my code. The `remove_account()` function accepts an account dict object, not an account name.

dmulder:
Make test works now! Just need to wait on the kanidm MRs to merge, and this one is done!

dmulder:
Getting a join error:
```
Dec 21 15:25:04 localhost.localdomain himmelblaud[4790]: 00000000-0000-0000-0000-000000000000 ERROR    🚨 [error]: Failed to join domain: Failed to write domain join configuration: Os { code: 30, kind: ReadOnlyFilesystem, message: "Read-only file system" }
```
This is because systemd doesn't let us write to the himmelblau.conf. I'll need to store the joined data somewhere else.

dmulder:
The default connection timeout is also still a little low (doesn't account for domain join time).

dmulder:
Everything is working now in testing. Just waiting on the kanidm MRs to merge.

PR #72, opened 2024-02-22
WIP: Use the SSSD id mappings

### LINKED ISSUE #71: Use SSSD idmapping
Use SSSDs ID mapping to ensure consistency. No reason to be re-inventing the wheel here.

**WARNING: This will be a breaking change and not backwards compatible with Himmelblau 0.1.x!**

dmulder:
Fixes #71

Checklist

- [ ] cargo fmt has been run


dmulder:
@Firstyear would you mind taking a look at how I've done the FFI bindings here? My main concern was the need to implement Send/Sync on the FFI. I wrapped the raw ctx pointer in a RWLock to make it safer for cross-thread use. I'm not sure if this is enough though.

src/idmap/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+        )
+    ))
+}
+
+pub const DEFAULT_IDMAP_RANGE: (u32, u32) = (200000, 2000200000);
+
+// The ctx is behind a read/write lock to make it 'safer' to Send/Sync.
+// Granted, dereferencing a raw pointer is still inherently unsafe.
+pub struct SssIdmap {
+    ctx: RwLock<*mut ffi::sss_idmap_ctx>,
```
dmulder, line 110: Right here is where I mean. That raw pointer was preventing me from calling this library from the idprovider.

dmulder: Hrm, still interested in your opinion, but it looks like this whole approach is just not going to work. SSSD idmapping depends heavily on SIDs. There are no SIDs in AAD. I can fake a SID for users/groups, but not for the domain. Then SID to DOM_SID matching doesn't work anyway. Sigh

Firstyear: I was actually going to caution you out of linking to SSS as a whole given it's history of "quality". SID mapping to an ID should be an algorithm we can easily copy if needed anyway. 

In this case, I think we need some other way to generate stable uids. If each user has a uuid you could take the last 32 bits as the uid/gid number perhaps? We do this in kanidm, but you do quickly approach the birthday problem ...

src/idmap/build.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    let bindings = bindgen::Builder::default()
+        .blocklist_function("qgcvt")
+        .blocklist_function("qgcvt_r")
+        .blocklist_function("qfcvt")
+        .blocklist_function("qfcvt_r")
+        .blocklist_function("qecvt")
+        .blocklist_function("qecvt_r")
+        .blocklist_function("strtold")
+        .header("wrapper.h")
```
Firstyear, line 16: To prevent over-rebuilding you want rerun-if-changed on OUT_DIR, bindings.rs and build.rs 

https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection

src/idmap/src/lib.rs
Firstyear: > Right here is where I mean. That raw pointer was preventing me from calling this library from the idprovider.

I'm assuming because it caused the async to fail because you were no longer send/sync?

dmulder: > I'm assuming because it caused the async to fail because you were no longer send/sync?

Yeah, exactly.

dmulder: > In this case, I think we need some other way to generate stable uids. If each user has a uuid you could take the last 32 bits as the uid/gid number perhaps? We do this in kanidm, but you do quickly approach the birthday problem ...

This is something I thought of doing. I had hoped to exactly mirror the behavior of either SSSD or Winbind, but neither will really work well with these IDs vs SIDs. So I think we'll need to come up with our own similar approach.

Firstyear: Do we know how windows turns an azure ad uuid into a SID? That could be what we need to re-create actually.

src/idmap/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+        ffi::idmap_error_code_IDMAP_NAME_UNKNOWN => IdmapError::IDMAP_NAME_UNKNOWN,
+        ffi::idmap_error_code_IDMAP_ERR_LAST => IdmapError::IDMAP_ERR_LAST,
+        _ => {
+            error!("Unknown error code '{}'", e);
+            IdmapError::IDMAP_ERROR
+        }
+    }
+}
+
+fn object_id_to_sid(object_id: &Uuid) -> Result<String, IdmapError> {
```
dmulder, line 71: > Do we know how windows turns an azure ad uuid into a SID? That could be what we need to re-create actually.

This is how. I created this function based on Windows behavior when joined to azure ad.

Firstyear: From that could we re-implement the idmap SID -> UID process then in here?

dmulder: I believe so, and that's probably what I'll end up doing. I've taken a break on this one and I'm working on a more critical bug at the moment.

PR #102, opened 2024-04-16
Add rid idmapping (replacing existing idmap)

dmulder:
Fixes #

Checklist

- [x] This pr contains no AI generated code
- [x] cargo fmt has been run
- [x] cargo clippy has been run

### DIFF
--- a/src/common/Cargo.toml
+++ b/src/common/Cargo.toml
@@ -21,8 +21,6 @@ tracing = { workspace = true }
 configparser = "^3.0.2"
 msal = { workspace = true }
 graph = { workspace = true }
-rand = "^0.8"
-rand_chacha = "^0.3"
 himmelblau_policies = { workspace = true }
 reqwest = { workspace = true }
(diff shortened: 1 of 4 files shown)


dmulder:
This has the obvious problem of a higher likelihood of uid collisions. Unfortunately, the 'rid' of a generated SID isn't guaranteed to be unique, as in a on-prem domain.

dmulder:
> This has the obvious problem of a higher likelihood of uid collisions. Unfortunately, the 'rid' of a generated SID isn't guaranteed to be unique, as in a on-prem domain.

It's still an improvement over the current mapping attempt though. This will necessarily be a temporary solution, and is waiting on the Samba and SSSD communities to settle on a proper ID mapping standard.

dmulder:
Otherwise, I tested this, and it appears to resolve the nss issues which were happening previously. It seems that the previous id mapping attempt was inconsistent in providing ids.

Issue #103, opened 2024-04-17
Himmelblau doesn't handle SFA users


dmulder:
While MFA now works smoothly, SFA users (those who have not enabled MFA) are unable to sign in at all.
This is a show stopper for the 0.3.0 release, and needs to be fixed.

dmulder:
There is an additional problem here. The user is only SFA because it hasn't had MFA enrolled yet, but MFA is required. Fixing the deadlock in the DAG allows us to SFA the user, but then Windows Hello enrollment fails. Apparently Windows Hello only works with MFA.

dmulder:
Fixed by #108

PR #104, opened 2024-04-19
Fix fallback DAG deadlock

dmulder:
The DAG request isn't polling properly, and needs
patches to Kanidm. Disable DAG for now until this
is fixed.

Fixes #

Checklist

- [ ] cargo fmt has been run

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.3.0"
+version = "0.3.1"
 authors = [
     "David Mulder <dmulder@suse.com>"
 ]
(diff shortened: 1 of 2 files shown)


dmulder:
This is a temporary fix for https://github.com/himmelblau-idm/himmelblau/issues/103. A more permanent fix will require a redesign of the DAG in Kanidm.

dmulder:
Perhaps this fix should remain anyway, and we could have a fallback from the SFA fail to DAG (when that's fixed). That way we have a shortcut to enrollment if the user doesn't require MFA anyway (otherwise they'll be forced to SFA authenticate *again* using the DAG).

dmulder:
Updated to include the DAG fallback after SFA attempt.

PR #203, opened 2024-10-03
Fixes for the dbus broker

dmulder:
This resolves the himmelblaud hang caused by
async/block_on issues. The broker now
communicates with himmelblaud via a unix socket,
instead of an additional dbus service.

Fixes #

Checklist

- [ ] cargo fmt has been run


dmulder:
After testing, this is a significant improvement, but there are some bugs that still need some work. I'll fix these in a follow up MR.

PR #335, opened 2025-01-03
Rewrite the sso code in Rust

dmulder:
This rewrites the Python3 SSO code in Rust.


dmulder:
This doesn't actually work yet. In testing, the Rust binary behaves exactly the same in the terminal, but the SSO browser plugin terminates the connection for some reason.

fmoessbauer:
> This doesn't actually work yet. In testing, the Rust binary behaves exactly the same in the terminal, but the SSO browser plugin terminates the connection for some reason.

I'm pretty sure this is related to the line-format that is passed via stdin / stdout. The native messaging protocol is VERY picky about the data that is transferred.

Urist-McGit:
> This doesn't actually work yet. In testing, the Rust binary behaves exactly the same in the terminal, but the SSO browser plugin terminates the connection for some reason.

I have tested your implementation on my local machine in Firefox, and it worked fine for me. The native messages also match bit by bit as far as I have seen. Maybe some mistakes were made in your test setup?

src/sso/src/main.rs
```diff
@@ -0,0 +1,316 @@
...
+        io::stdin().read_exact(&mut buffer)?;
+
+        let message_str = String::from_utf8(buffer)?;
+        Ok(serde_json::from_str(&message_str)?)
+    }
+
+    fn encode_message<T: Serialize>(message_content: &T) -> Result<Vec<u8>, Box<dyn Error>> {
+        let encoded_content = serde_json::to_vec(message_content)?;
+        let mut result = Vec::new();
+        result.extend(&(encoded_content.len() as u32).to_le_bytes());
```
Urist-McGit: I think the specification says native endianness. The python app uses `struct.pack("@I", ...)` here, so `to_ne_bytes` should be correct.

Urist-McGit (commented):
I also had a build failure, the `new_v4` was not available from the `uuid` crate. I needed to enable the `v4` feature for it.

src/sso/src/main.rs
dmulder: Ah! That makes sense. I'll test this.

dmulder: Well, I've tried modifying that line and using native endianness, but it still fails the same way.
Screenshot From 2025-01-07 14-26-45

dmulder:
> native messaging protocol

Perhaps I should be using a [native messaging library](https://crates.io/crates/native_messaging), rather than re-inventing the wheel.

src/sso/src/main.rs
Urist-McGit: You could create a breakpoint in the JS background task, there should be a function named `on_native_message`. From there you could inspect the native message sent by your implementation and see if something is wrong with it.

dmulder: Screenshot From 2025-01-08 08-34-15

So something is wrong with how I'm sending the message. Perhaps the bytes count is wrong?

src/sso/src/main.rs
```diff
@@ -0,0 +1,316 @@
...
+            println!("{}", serde_json::to_string_pretty(&token)?);
+        }
+        _ => Args::command().print_help()?,
+    }
+    Ok(())
+}
+
+async fn run_as_native_messaging() -> Result<(), Box<dyn Error>> {
+    println!("Running as native messaging instance.");
+    println!("For interactive mode, start with --interactive");
```
dmulder: (facepalm) These messages need to go to stderr, otherwise they're interpreted as corrupted native messages.

dmulder:
> > native messaging protocol
> 
> Perhaps I should be using a [native messaging library](https://crates.io/crates/native_messaging), rather than re-inventing the wheel.

I tried using this existing native messaging library, but it does not work. It sent the message wrapped in quotes, and the browser plugin couldn't parse it. I ended up getting my original implementation to work, so I'll stick with it.

src/sso/src/main.rs
Urist-McGit: Funnily enough I fixed that to compare the messages on stdout. That is why it worked for me!

dmulder: Lol! That explains why you got it working when I couldn't. It was just a simple mistake.

PR #344, opened 2025-01-16
Support password changes when demanded

### LINKED ISSUE #342: Implement Entra Id password changes
**Summary**
Right now if a new user hasn't yet set their password, and if they're logging in for the first time using an admin provided generated password, we are unable to handle the request, and recommend a DAG. We should simply handle this request and allow them to set the password.

**Problem Statement**
New users logging in with a generated password face difficulties as Himmelblau does not support in-session password changes. This limitation requires a workaround, such as using a separate device.

**Proposed Solution**
Implement password changes that can happen during the authentication flow. Most of the change will happen in libhimmelblau.

**Benefits**
Simplifies new user setup.

**Alternatives Considered**
None.

**Additional Context**
https://github.com/Gerenios/AADInternals/blob/b23a7845f6dc5ea8c57b10351421a4d00466cd90/AccessToken_utils.ps1#L2336

dmulder:
At times it is necessary to provide a password
reset flow, such as first-time-setup (when the
user receives a temp password from the admin),
and when the user's password is reset.

Fixes #342


dmulder:
This still needs a little work. It now supports password changes for password-only users, but it has a hickup when the password is reset for an MFA user. For an MFA user, the password reset succeeds, but then the auth seems to fail (with seemingly no logged error). Immediately following that failed auth, the user can then auth with the new password and the auth succeeds. So the flow works, but is clunky at the moment. So it still needs work.

dmulder:
The remaining issue is because the MFAPoll is called from a macro in the pam module. The macro can't call itself, so it isn't nested, and multiple MFA prompts in a single interaction aren't currently possible. I'll need to find a way to resolve this.

dmulder:
I think these latest patches resolve the remaing MFAPoll issue. Needs testing.

PR #355, opened 2025-01-30
Use posix attributes synchronized from on-prem AD

### LINKED ISSUE #341: Utilize on-prem synched scheme extensions for gecos/gidNumber/uidNumber/unixHomeDirectory/loginShell
**Summary**
Enable Himmelblau to support custom POSIX attributes stored in Entra ID by leveraging the Entra schema extensions and the Microsoft Graph API.

**Problem Statement**
Currently, Himmelblau maps the `uid` and `gid` attributes to the UPN of the user to ensure consistent mapping and proper SSH functionality. The `gecos` field is retrieved from the OAuth2 token received after authentication, while the home directory and login shell are configured in the `himmelblau.conf` file. This fails to account for current on-prem Active Directory user attributes. Without synchronization, file permissions become inconsistent when transitioning between Entra ID and Active Directory environments, leading to issues with cross-environment file access.

**Proposed Solution**
Introduce the ability for Himmelblau to retrieve custom POSIX attributes from Entra ID using schema extensions accessible via the Microsoft Graph API. The approach would involve:

1. Continue using the current UID/GID mapping against the UPN but correct it to use the schema-mapped attributes once they are detected during authentication.
2. Add a command to `aad-tool` that authenticates a user capable of reading from the Graph and caches current users and groups with their respective POSIX attributes.
3. Provide an optional integration where admins register an app in Azure with Graph access and distribute a client secret for querying POSIX attributes.

Combining options 1 and 2 would provide a balance between ease of configuration and reliable attribute resolution. Option 3 can be offered for finer control where required.

**Benefits**
- Ensures proper UID/GID synchronization with on-prem Active Directory.
- Enables POSIX attribute synchronization directly from Entra ID.
- Offers both minimal and advanced configuration paths, suiting various deployment needs.
- Reduces reliance on user principal name (UPN)-based UID/GID assignment, avoiding issues when UPN changes.

**Alternatives Considered**
Setup an admin-configurable service account to query Entra ID for UID/GID information and keep attributes current across hosts.

**Additional Context**
- Microsoft documentation on schema extensions: [Directory Extensions in Entra ID](https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/how-to-connect-sync-feature-directory-extensions)
- Previous considerations included device credentials, but Graph API access requires user tokens.

dmulder:
Fixes #341

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ tracing-subscriber = "^0.3.17"
 tracing = "^0.1.37"
 himmelblau_unix_common = { path = "src/common" }
 kanidm_unix_common = { path = "src/glue" }
-libhimmelblau = { version = "0.5.7", features = ["broker", "changepassword"] }
+libhimmelblau = { version = "0.5.8", features = ["broker", "changepassword"] }
 clap = { version = "^4.5", features = ["derive", "env"] }
 clap_complete = "^4.4.1"
 reqwest = { version = "^0.12.2", features = ["json"] }
--- a/man/man5/himmelblau.conf.5
+++ b/man/man5/himmelblau.conf.5
@@ -167,6 +167,14 @@ A boolean option that enables the experimental multi-factor authentication (MFA)
 enable_experimental_mfa = true
 
 .TP
+.B sync_unix_schema_extension_attributes
+.RE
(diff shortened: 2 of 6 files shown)


dmulder:
This needs testing before it will merge.

dmulder:
This is working for user posix attrs, but it fails for groups.

dmulder:
> This is working for user posix attrs, but it fails for groups.

Sigh, this is the same issue we've been facing with group names. MS is rejecting the request because the user only has the rights to read the group id, and nothing else.

dmulder:
The problem with this group read issue, is if an admin and non-admin authenticate to the same host, the gid for the group will switch back and forth in nss, depending on whether the user (the group member) has rights to read the property.

PR #366, opened 2025-02-13
Add Nix build definitions and a rudimentary NixOS module

twoolie:
This PR adds support for building himmelblau with Nix, and a fairly minimal NixOS module for inclusion into a NixOS system configuration. It is structured as a [flake](https://nix.dev/concepts/flakes.html#flakes), which is very well supported despite it's experimental label and allows me to define everything needed in one file (plus a lockfile).

The package definition currently only includes the WebKit dependencies if the "interactive" feature is enabled under the assumption that this is what is meant by interactive 2FA login.

The NixOS module is very simple, and only allows configuring a single domain for now. I hope to split the options apart when I figure out which options make sense to override per-domain and which are purely global.

There is some basic documentation including usage examples showing how to include the module in your own NixOS configuration. I have not tried to document all the options in the readme, although they will be automatically included in the man pages of the built NixOS system.

There is an entry in the Makefile to enable building the nix package. this currently relies on a native nix install, but it could be modified to use a persistent nix docker container if needed.

I have added simple sdnotify readiness reporting, as I had observed the tasks daemon starting up too soon and looping waiting for the socket to become available, which can take a while. This can also improve the login experience by preventing login before himmelblaud is ready to accept connections.

I have also added a simple CI job to check the validity of the flake, and build the two variants of the package.

Please let me know if there are any changes you would like to see, or clarifications needed.


dmulder:
:+1: Awesome work @twoolie! Unfortunately, the nixos CI is failing. Could you look into this?

I'm also modifying the README installation instructions, so you might need to rebase (depending on whose MR is ready first).

twoolie:
I'm not sure why nix is failing to clean up a source folder, I have not seen this happen in my testing. hopefully it's a transient problem with the runner.

twoolie:
I've rebased against main. I had to edit one of the tests to call a locally provided test script as /bin/echo is not available in the nix build environment.

PR #429, opened 2025-03-11
Configure Cachix signing and upload in CI

twoolie:
This adds the minimal config to sign and push build derivations to Cachix.
Included is documentation for how to configure your client to pull and verify built packages.

You will need to create a (free-tier) account at https://app.cachix.org/signup (i'm assuming you pick the name `himmelblau`) and generate some signing keys locally with:

```sh
$ cachix authtoken <your auth token>
$ cachix generate-keypair himmelblau
```

The signing key is printed to stdout, and this is the **only copy** so make sure you [copy it straight into github](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository) with the name `CACHIX_SIGNING_KEY`.

Please let me know if you have any issues setting this up, or if you would like some further documentation written.


dmulder:
I have no idea how to install cachix. I looked around, and it seems I need the nix command to install cachix (even to build it), which I also don't have and don't know how to install.

twoolie:
You will need to have the nix tool installed locally. It can be installed on any linux or darwin system, and will co-exist with your existing package manager. The easiest way is to use this rust-based installer from determinate systems:

```sh
$ curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
```

After you have reloaded your session (to get the new paths on your PATH), you can install the cachix client with `nix profile install --accept-flake-config 'nixpkgs#cachix'`.

If you don't want to or can't get it working, you can add me as a contributor and I can set this up on your behalf.

dmulder:
I made the mistake of running `cachix generate-keypair himmelblau` before creating the cache at `https://app.cachix.org/cache/himmelblau`. It threw this error:

Binary cache himmelblau doesn't exist or you're not authorized to access. Please check the name for typos and/or provide an auth token. Contact support on https://cachix.org if you're stuck!

But since I ran that too early, now doing a `cachix generate-keypair himmelblau` throws this error:

A public signing key already exists. It's currently not possible to override or add multiple signing keys.

That looks like a bug in cachix to me. Don't store the key pair locally if it failed to upload. @twoolie do you know how to fix this?

twoolie:
What version of cachix did nix install for you? I can't seem to replicate this bug on any of the more recent versions.

As an immediate fix, you can edit `~/.config/cachix/cachix.dhall` to remove the public key that exists for `himmelblau`.

dmulder:
> What version of cachix did nix install for you? I can't seem to replicate this bug on any of the more recent versions.

cachix 1.7.6

dmulder:
My `~/.config/cachix/cachix.dhall` doesn't have a public key, just an auth token.

twoolie:
Ah, that's strange. The only other place I can think it would be hiding is
in ~/.config/nix/nix.conf

I took a look through the cachix source and it looked like it was only
checking cachix.dhall for existing keys.

twoolie:
I couldn't find the text "A public signing key already exists" anywhere in the client code, so i went looking through the [generate-keypair code here](https://github.com/cachix/cachix/blob/v1.7.6/cachix/src/Cachix/Client/Command/Config.hs#L36-L51), the order of operations is:

* Check that you have an auth token (`getAuthTokenRequired`)
* Generate a Keypair (`createKeypair`) with `pk` and `sk`.
* Call the API with the auth token and public key, checking if you are allowed to create one. 
   * This would be where you hit your first error
* Add the signing key to your local cachix.dhall and save it

I think it's safe to say that the error is not client-side, have you checked the web interface at [app.cachix.org](https://app.cachix.org) to see if they have a public key on that side that can be deleted?

dmulder:
> I think it's safe to say that the error is not client-side, have you checked the web interface at [app.cachix.org](https://app.cachix.org) to see if they have a public key on that side that can be deleted?

I created a `himmelblau` cache on at [app.cachix.org](https://app.cachix.org), and it shows it has an available public key. I don't see a way to delete it, other than deleting the whole cache. Besides, I already tried deleting that and re-creating it, but I still get the same error.

dmulder:
You have to select 'Self-signed' for the signing key. You also never mentioned that I needed to create the cache on cachix.

Screenshot From 2025-03-18 08-47-07

PR #464, opened 2025-04-03
WIP: Add Oracle Linux 7.9 build

dmulder:
This adds support for RHEL 7.9 et al, but with some caveats. As with RHEL 8 et al, Hello doesn't work (openssl is too old). But additionally, we can't build the sssd idmap code in 7.9, because utf8proc is too old also (and the toolchain is old). This means we're limited to no Hello, and idmap options of UUID and RFC2307.


dmulder:
Cancelling this for now, unless someone else really needs it.

PR #479, opened 2025-04-16
Fix Nix Flake

iLikeToCode:
The package errors out with
```lib.fileset.gitTracked: Expected the argument (...) to point to a local working tree of a Git repository, but it’s not.```
as nix pulls without the .git folder.
Changing src to ./. fixes this problem and allows the package to build.

Tested in Hyper-V on NixOS 24.11

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,7 @@
           recipe = {lib, enableInteractive ? false}: rustPlatform.buildRustPackage {
             pname = "himmelblau";
             version = cargoToml.workspace.package.version;
-            src = ./.
+            src = ./.;
             outputs = [ "out" "man" ];
             cargoLock = {
               lockFile = ./Cargo.lock;
--- a/flake.nix
+++ b/flake.nix
@@ -12,11 +12,7 @@
           recipe = {lib, enableInteractive ? false}: rustPlatform.buildRustPackage {
             pname = "himmelblau";
             version = cargoToml.workspace.package.version;
-            src = with lib.fileset; toSource {
-              root = ./.;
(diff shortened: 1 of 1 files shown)


iLikeToCode:
This is not confirming that the package works, but that change only affects build time, not runtime, but it builds now, didn't before.

iLikeToCode:
Curious why the build failed when that task doesn't build on nix...

dmulder:
> Curious why the build failed when that task doesn't build on nix...

It's unrelated:
https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts

I'm trying to figure out how to migrate now.

iLikeToCode:
As long as the nixos goes through, which it should, that is all that commit affects, it built infront of my eyes 10 minutes ago so *should* be good

iLikeToCode:
As I said, package may or may not work, I am connecting to my high schools entra id (with permission) but yk what the policies can be like on education AD

dmulder:
@twoolie could you review this change?

twoolie:
> nix pulls without the .git folder.

I don't think this is true for flakes, but i'm open to being wrong here. I've been pulling this repo into our server's flake for a while now, and it builds fine for us.

@iLikeToCode Can you tell me more about how you're pulling and building the project?

Either way, the parts you deleted are actually needed for development, otherwise every change you make in the repo forces a re-compile. I'd rather fix the problematic gitTracked filter than delete the whole lot.

dmulder:
I'm going to reject this request for the time being. It sounds like this needs to be solved in some other way. @twoolie can you spearhead fixing this issue?

PR #519, opened 2025-05-19
Order services before gdb/nss-user-target

Firstyear:
Ensure that the himmelblau resolver has started before the nss-user target. This target is very early in startup, and ensures that users can be resolved for *any* graphical display manager or incoming ssh request.


dmulder:
Tasks socket is at /var/run/himmelblaud/task_sock

PR #580, opened 2025-06-25
Support hardware TPM

dmulder:


dmulder:
@twoolie It looks like NixOS uses too old of a cargo version to build this new code. I'm merging it into main anyway, but we need to figure out a solution prior to the 1.0 release. The minimum version is now cargo 1.85.0, FYI.

dmulder:
@twoolie FYI on some distros I'm installing cargo via `rustup` to ensure compatibility. Perhaps we can do that in NixOS?

PR #590, opened 2025-07-01
Enroll in Intune as soon as it is enabled

### LINKED ISSUE #551: Intune enrollment needs to happen for upgraded devices
**Description**
Right now Intune enrollment only happens when the device is enrolled in Entra Id. This means Intune enrollment will be skipped for devices upgraded from <=0.9.x. We should also check for the case where the device is Entra Id enrolled, but not Intune enrolled, then complete the enrollment.

---

### 💡 Help Make It Happen!
Want to see this bug fixed faster? Fund its implementation through our **Backer's Bounty** program, where you choose which bug fixes get priority!

dmulder:
This will allow upgrading a device and enabling policy enforcement. Previously the Intune enrollment was only happening at device join time.

Fixes #551


dmulder:
There is a problem with this solution. We cannot request an Intune enrollment token using a Hello key. The only way around this is to force an MFA.

PR #593, opened 2025-07-07
Himmelblau should get posix attributes w/out delegate user access

### LINKED ISSUE #401: himmelblau Entra ID App should get posix attributes without delegate user access (rfc2307)
**Summary**
Currently, the first login does not work as expected because the home directory is created with a fake uid/gid. Therefore, we have to log in a second time via gui or ssh to get a fully functional desktop. This behaviour should be corrected.

**Problem Statement**
At the first login, himmelblau creates a fake uid/gid because it can only determine the correct uid/gid with rfc2307 attribute mapping after authenticating to the Entra ID. Some features of gdm do not work, such as search. The desktop  works properly  on the 2nd login.
The problem also occurs with ssh login. The first login attempt fails. The second login is successful because then the home directory has been created and have the right uid/gid attributes.

**Proposed Solution**
The himmelblau Entra app itself should get posix attributes from the Entra ID directory extension schema. 
For this proposal we need a client secret and other permissions. 

GroupMember.Read.All Type: Application
User.Read.All Type: Application

This allows himmelblau to get the groups and users with a uid/gid independent of a required user login.

Of course, the access token for the user login must be checked. 
And the stored client secret must be protected.

**Benefits**
The user has a better experience because there is no confusion about the first login and all works like expected.

dmulder:
Fixes #401


dmulder:
This is now working, but there are some lingering gaps that need resolved.
For example, the aad-tool command should utilize the client credentials if available (and the right client_id was selected).
Also, we can now implement group lookups for the daemon. Currently, we only know about group memberships, and can't lookup groups on the fly. This will only work by object id, but still potentially useful.

PR #623, opened 2025-07-16
Handle disabled device by attempting re-enrollment

### LINKED ISSUE #151: Deleting device object prevents authentication
We need to add an 'unenroll' option (perhaps to aad-tool?), otherwise Himmelblau just fails to authenticate when the device object is deleted from the directory. 
Right now the only work around is to delete everything in /var/cache/himmelblaud/ (and maybe /var/lib/himmelblaud/hsm-pin also?).

dmulder:
Fixes #151

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,16 +20,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1739206421,
-        "narHash": "sha256-PwQASeL2cGVmrtQYlrBur0U20Xy07uSWVnFup2PHnDs=",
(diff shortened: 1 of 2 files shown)


dmulder:
The purpose of this change is to recognize when a device has been disabled/removed in the directory. If this is the case, we now automatically delete the enrollment keys, therefore forcing a new enrollment.
Obviously, if the device was disabled inadvertently, etc, we re-enroll. If the device was disabled on purpose (perhaps the machine was stolen), we are now dead in the water, and our keys have been deleted.

PR #627, opened 2025-07-18
Add `fedora42` build target

s8n-no:
Adds `fedora42` as a build target, corresponding container image, and add `utf8proc-devel` as a build dependency.

The build targeting `rawhide` doesn't install on Fedora 42, as `pam-himmelblau` will depend on a newer version of libFLAC.

I've tested Himmelblau on multiple Fedora 42 machines since it's release, and it's works well.

### DIFF
--- a/images/rpm/Dockerfile.fedora42
+++ b/images/rpm/Dockerfile.fedora42
@@ -1,0 +1,51 @@
+# Use the official Fedora 42 image as the base
+FROM fedora:42
+
+# Update the package list and install essential build dependencies
+RUN dnf -y update && \
+    dnf -y install \
+    wget \
+    git \
+    gcc \
+    gcc-c++ \
+    make \
+    openssl-devel \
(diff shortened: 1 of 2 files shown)


dmulder:
FYI, `utf8proc-devel` is no longer a build dependency (this was changed rather recently).

images/rpm/Dockerfile.fedora42
```diff
@@ -0,0 +1,51 @@
...
+    gobject-introspection-devel \
+    cairo-devel \
+    libsoup3-devel \
+    javascriptcoregtk4.1-devel \
+    webkit2gtk4.1-devel \
+    pango-devel \
+    atk-devel \
+    gtk3-devel \
+    tpm2-tss-devel \
+    utf8proc-devel \
```
dmulder:
```suggestion
```

This dependency is no longer required.

PR #631, opened 2025-07-22
Additional Intune policy enforcements

dmulder:


dmulder:
The password policies work, but the custom compliance does not. It appears we may be missing a step somewhere for custom compliance.

dmulder:
This marks custom compliance policies as experimental. Hopefully we can stablize the behavior of custom compliance prior to 2.x.

PR #644, opened 2025-08-06
Automatically append to nsswitch.conf in postinst

iLikeToCode:
Automatically append to nsswitch.conf in postinst
Not tested yet but I'll test it shortly

### DIFF
--- a/src/daemon/scripts/postinst
+++ b/src/daemon/scripts/postinst
@@ -15,5 +15,9 @@ if grep -q '^ID=debian' /etc/os-release && grep -q '^VERSION_ID="12"' /etc/os-re
     systemctl daemon-reload || true
 fi
 
+sed -i '/^passwd:/ {/himmelblau/! s/\<files\>/& himmelblau/}' /etc/nsswitch.conf
+sed -i '/^group:/ {/himmelblau/! s/\<files\>/& himmelblau/}' /etc/nsswitch.conf
+sed -i '/^shadow:/ {/himmelblau/! s/\<files\>/& himmelblau/}' /etc/nsswitch.conf
+
 # Add nss cache directory with correct permissions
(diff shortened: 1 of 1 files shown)


dmulder:
Your patch actually doesn't work, for a couple of reasons.

First, not all nsswitch.conf lines contain the word 'files'. For example, here is my nsswitch.conf (on openSUSE Tumbleweed):

```conf
passwd:		compat systemd himmelblau
group:		compat [SUCCESS=merge] systemd himmelblau
shadow:		compat systemd himmelblau
```

Second, the nsswitch.conf has been moved to `/usr/etc/nsswitch.conf` on some newer distros.

I recommend you first check if `/etc/nsswitch.conf` exists before attempting to apply the patch, then check if `/usr/etc/nsswitch.conf` exists and attempt to apply the patch there. If we can't find an nsswitch.conf, we'll ignore it for now.

Next, these commands I think will work better:

```
sed -i '/^passwd:/ {/himmelblau/! s/$/ himmelblau/}' /etc/nsswitch.conf
sed -i '/^group:/ {/himmelblau/! s/$/ himmelblau/}' /etc/nsswitch.conf
sed -i '/^shadow:/ {/himmelblau/! s/$/ himmelblau/}' /etc/nsswitch.conf
```

Obviously while accounting for the different nsswitch.conf locations.

dmulder:
One more thing, these should go in the nss-himmelblau package, not himmelblau.

I'm reworking your patch now and will merge shortly.

iLikeToCode:
ok thanks for the feedback

dmulder:
I've reworked the patch here: https://github.com/himmelblau-idm/himmelblau/pull/646/commits/9178044086c33e48712041f2e1e3322e87b404af

iLikeToCode:
thank you

dmulder:
Merged with #646

PR #648, opened 2025-08-08
cron dep for Intune and SELINUX config

dmulder:
Additionally add various build improvements.


dmulder:
@twoolie This change breaks the NixOS build, but I have no idea how to resolve it. I'm going to leave it broken on the main branch for now, so I hope you can find time to look into this. I tried to simply ignore the selinux build in the nixos/ci, but it doesn't pick up the environment variable.

PR #723, opened 2025-09-21
feat: Add support for aarch64 on Debian-based distributions

### LINKED ISSUE #717: Device Auth Failure on ARM
**Description**
I have compiled the code for arm64 (by downloading the source into an arm64 machine and running the make commands after installing docker).  When running aad-tool auth-test on a user in the domain, the logs show that the intune enrolled check is conducted but is not acted upon.  The device does not appear in the enrolled devices list in Entra.  This issue does not occur in the x86_64/amd64 version.

**Steps to Reproduce**
1.  Compile himmelblau on arm64 for ubuntu 24.04
2.  Configure himmelblau
3.  Attempt to login

**Logs and Output**
Auth check components of the arm64 and x86_64 / amd64 logs are here:
[ARM Issue.txt](https://github.com/user-attachments/files/22389630/ARM.Issue.txt)

**Configuration**
global section of /etc/himmelblau/himmelblau.conf:
```
[global]
local_groups = users    # default local_groups (additional)
home_attr = CN          # UUID, SPN, CN # MUST match home_alias
home_alias = CN         # UUID, SPN, CN # MUST match home_attr
enable_hello = false    # windows hello
home_prefix = \home\\
use_etc_skel = true
domains = $domain       # domains from which to allow connections / authentication
pam_allow_groups = $EntraIDGroups # groups to allow (Object ID from Entra)
apply_policy = true
```

**Packet Trace (For Authentication Errors)**
The GET call with scope "openid profile" to the enrollment resource is present
The subsequent POST command includes the flowToken, username, and authentication option acceptance booleans.
The next (and last) GET call of note is to the portal at azure and has a size of >500kb.
If additional details are needed, I can likely provide them.

**Environment**
- **Linux Distro**:  Ubuntu 24.04 LTS
- **Package source (distro package/github release/self built)**: self built (as no arm64 in releases)
- **Himmelblau Version**: 1.2.3 (and 1.2.3-dev with mitm proxy flags added)

**Additional Information**
I have tested this exact same configuration script on an x86_64 / amd 64 system (using the files from release) and everything worked as expected.
I am not well versed in this build process that utilizes docker so if the basic steps of running the make command on an arm64 system after having installed docker is insufficient, that could be the issue.

RobertDeRose:
This updates the Debian asset locations for the `libnss` and `libpam` modules to use `cargo-deb`’s `--multiarch=same` option, ensuring that shared libraries are installed in the correct architecture-specific subdirectory on Debian-based distributions.

The `gen_dockerfiles.py` script has also been updated to:
* Use the JSON-formatted CMD structure
* Format the output across multiple lines for readability
* For Debian-based distributions, adds the `--multiarch=same` flag for the `libnss` and `libpam` modules.

Fixes #717


src/common/src/idprovider/himmelblau.rs
```diff
@@ -49,7 +49,6 @@ use kanidm_hsm_crypto::{
 use regex::Regex;
 use reqwest;
 use reqwest::Url;
-use serde::{Deserialize, Serialize};
```
RobertDeRose, line 52: I had to remove these lines and the `struct` below to get the workspace to build, otherwise I would not have touched this file.

RobertDeRose:
@dmulder I saw the comments made on #717 and it took me a lot of time to figure out that it was just a simple matter of the libraries being installed in the correct location on arm64 distress. 

I was able to reproduce the same issue as the author of 717, but with this patch, I a get the MFA request on my phone. The only issue I personally have seem to be configuration related with either Himmelblau or Entra, but I am not sure, as I am not an expert of either.

```json
{"Denied":"AADSTS50001: The service principal for resource 'https://enrollment.manage.microsoft.com/' is disabled. This indicate that a subscription within the tenant has lapsed, or that the administrator for this tenant has disabled the application, preventing tokens from being issued for it. Trace ID: 0ac490d5-b561-4fc9-99fa-e4a1bc482100 Correlation ID: 20bd8b9f-c949-454f-a869-7a7a22073300 Timestamp: 2025-09-21 02:59:34Z"}}
```

From my research, it seems to be unrelated to Himmelblau, unless I can disable the device enrollment process, which might solve my issue for the time being, just have to figure out if I can or not.

src/common/src/idprovider/himmelblau.rs
dmulder: What error were you getting when you tried to build with these?

dmulder: Huh, it's not even used anywhere. I don't remember what this was here for. Cargo should have caught that dead code on x86 also.

dmulder:
> From my research, it seems to be unrelated to Himmelblau, unless I can disable the device enrollment process, which might solve my issue for the time being, just have to figure out if I can or not.

It's saying you don't have the permission to enroll the device in Entra Id. Your Entra admin has probably disabled device enrollment except for specific IT admin users. Unfortunately, Himmelblau currently requires device enrollment (though I'm considering providing a non-enrolled option).

RobertDeRose:
> > From my research, it seems to be unrelated to Himmelblau, unless I can disable the device enrollment process, which might solve my issue for the time being, just have to figure out if I can or not.
> 
> It's saying you don't have the permission to enroll the device in Entra Id. Your Entra admin has probably disabled device enrollment except for specific IT admin users. Unfortunately, Himmelblau currently requires device enrollment (though I'm considering providing a non-enrolled option).

You don't have to consider it, I have PRs waiting to be pushed up. I'm just testing some changes I need to get PAM to work with Cockpit and then you'll see the PRs. Hopefully by the end of this week if I can figure out why the cockpit-session runtime thinks the PAM conversation is invalid.

PR #780, opened 2025-10-10
DRAFT: Add better error messages

twoolie:
Motivated by a particularly unhelpful error message:
> `00000000-0000-0000-0000-000000000000 ERROR    :rotating_light: [error]:  | err: NotFound`

This PR aims to add better error messages, and surface those errors back to clients when possible.

The first error I have tackled is NotFound. this error message was generated in a lot of places, and contained no context as to what wasn't found, and where we were looking for it. `what` and `where_` have been added as fields, filled out as appropriate, and formatted into the final error message sent to clients and logs.

I also did a bit of refactoring to turn some of the match pyramids into straight-line error handling via `find_provider` and `idp_get_domain_for_account` returning `Result`s.

Next, I would want to tackle `BadRequest` (why was the request bad?) and then Transport. Eventually I would want all `IdpError` variants to have a field or two to carry specifics of their error conditions.

@dmulder Would you please take a look at this draft and give some feedback if this is going in the right direction?

### DIFF
--- a/src/common/src/idprovider/himmelblau.rs
+++ b/src/common/src/idprovider/himmelblau.rs
@@ -270,7 +270,7 @@ macro_rules! find_provider {
 }
 
 fn idp_get_domain_for_account(account_id: &str) -> Result<&str, IdpError> {
-    match split_username(&account_id) {
+    match split_username(account_id) {
         Some((_sam, domain)) => Ok(domain),
         None => {
             debug!("Authentication ignored for local user");
@@ -381,7 +381,7 @@ impl IdProvider for HimmelblauMultiProvider {
         tpm: &mut tpm::provider::BoxedDynTpm,
         machine_key: &tpm::structures::StorageKey,
     ) -> Result<bool, IdpError> {
-        let domain = idp_get_domain_for_account(&account_id)?;
+        let domain = idp_get_domain_for_account(account_id)?;
         let providers = self.providers.read().await;
         let provider = find_provider!(self, providers, domain)?;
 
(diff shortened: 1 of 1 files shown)


dmulder:
I love this! Thanks for putting in the work @twoolie. I'm fine with the way you've implemented this so far. One question, I assume `where_` has the underscore because it collides with a built-in token?

twoolie:
Yes, that's why the underscore is there. I'll continue adding error messages and refining the current implementation.

src/common/src/idprovider/himmelblau.rs
```diff
@@ -3128,7 +3055,8 @@ impl HimmelblauProvider {
                             ?e,
                             "Device auth failed for Intune device enrollment, delaying enrollment."
                         );
-                        return Err(IdpError::NotFound);
+                        // TODO: Is NotFound the correct error type here? BadRequest better here?
```
twoolie, line 3199: @dmulder Thoughts? Is `NotFound` a valid error here, or should we add an error code for request/enrollment failure?

dmulder: See the response handler for this call in `join_domain`. We're intentionally ignoring Intune Enrollment failures and logging the error. We **could** return some other type (`NotFound` does seem a bit odd here), but notice that when we return `NotFound`, the result is ignored and we proceed? Sometimes the Intune Enrollment fails the first time (immediately after Entra Id enrollment). The idea here is to temporarily ignore that failure.

dmulder: My thinking here was that `NotFound` in this instance means the 'device creds are not found', because the enrollment is new.

dmulder:
@twoolie FYI, I had a user test your patches in #801, and they are proving to be very useful. It made the problem much clearer.

dmulder:
I definitely want these patches in 2.0.

twoolie:
In that case, do you want to merge it as-is and I'll start a new PR for the
other error messages?

dmulder:
Up to you. We still have some time to get it into 2.0. It's still in alpha. I'm also thinking I'd like your debug improvements backported to 1.x, since they've proven very useful.

dmulder:
I'm going to merge this MR, since I'm going to push the 2.0 release soon. Feel free to post a follow up MR.

Issue #619, opened 2025-07-16
Add support for choosing alternate MFA methods


surry:
**Summary**

Enhance himmelblau and libhimmelblau to allow selecting alternate MFA options approved for Entra accounts. This is helpful in cases where the default/strongest MFA method is not available, but the organization has allowed other MFA methods in Azure.

**Problem Statement**

himmelblau currently seems to only allow the default MFA option for a user in Entra. A user is unable to select an alternate MFA option such as calling an approved office phone number when authenticating. 

**Proposed Solution**

Per @dmulder on linked discussion:
> For this particular issue, you first need to modify initiate_acquire_token_by_mfa_flow to respond with all available MFA options in [libhimmelblau](https://gitlab.com/samba-team/libhimmelblau). That will require modifying the MFAAuthContinue to hold multiple mfa_method types, instead of just the default. This should be relatively easy. We need to preserve some basic backwards compatibility (for security backports to previous releases of Himmelblau), so maybe add a new value mfa_methods to the MFAAuthContinue, and still set mfa_method to the default.
Secondly, Himmelblau itself will need to be modified to recognize the additional options. We'll need to modify the pam module to be able to prompt the user for the desired MFA method. We should probably provide a himmelblau.conf option for enabling/disabling this behavior.

**Benefits**
Users would be able to use himmelblau to authenticate with additional MFA options normally supported by Entra in a browser. e.g. the "I can't use my Microsoft Authenticator" here:

<img width="322" height="72" alt="Image" src="https://github.com/user-attachments/assets/e168a749-6522-4c9a-b497-8fc8054e7e66" />

<img width="226" height="278" alt="Image" src="https://github.com/user-attachments/assets/8871af15-42a6-409f-93ad-3f65f03d91f5" />

**Additional Context**
* https://github.com/himmelblau-idm/himmelblau/discussions/461#discussioncomment-12681766
---

### 💡 Help Make It Happen!
Want to see this feature developed faster? Fund its implementation through our **Backer's Bounty** program, where you choose which features get priority!

[![Donate to Our Collective](https://opencollective.com/himmelblau/donate/button.png?color=blue)](https://himmelblau-idm.org/backers.html#backers-bounty)

For US tax exempt donations:

[...]

PR #800, opened 2025-10-18
Add support for configuring a default MFA method.

surry:
# About

Implements support for configuring a default MFA method in himmelblau.conf as described in: https://github.com/himmelblau-idm/himmelblau/issues/619

The purpose of this is to allow configuration of the first MFA method used for authentication for cases where a user may not have access to the MFA device needed to authenticate. A user may have an alternate device/method like an approved office phone number that can be used for MFA instead.

The assumption here is that administrators installing himmelblau would only configure this option on systems where the default MFA method for a tenant cannot be used, and they are comfortable assuming the risk of an alternate, approved MFA method in Entra that may not be the strongest one.

# Considerations

* Depends on changes in libhimmelblau which are part of this MR: https://gitlab.com/samba-team/libhimmelblau/-/merge_requests/102

* The Docker build changes could be reverted if desired - I made these to make it easier to build himmelblau with an unpublished version of libhimmelblau from my MR using Cargo's patch functionality. There might be a better way to handle this that I missed, but I kept it in for now in case it would be helpful in the future.

* I used Claude Code to help generate some of the code and tests in here.

# Testing

I built a version of `himmelblau` with the changes from my `libhimmelblau` MR and installed it on an Ubuntu 24.04 EC2 instance. After modifying the `himmelblau.conf` with this line:

```ini
mfa_method = TwoWayVoiceMobile
```
I was able to login to my Entra account provisioned by himmelblau on this instance, with MFA using a voice call to a mobile phone, instead of my default authenticator app:

2025-10-18-himmelblau_working

### DIFF
--- a/src/cli/src/main.rs
+++ b/src/cli/src/main.rs
@@ -344,7 +344,13 @@ async fn auth(app: &BrokerClientApplication, account_id: &str) -> Option<UserTok
             let polling_interval = mfa_req.polling_interval.unwrap_or(5000);
             loop {
                 match app
-                    .acquire_token_by_mfa_flow(account_id, None, Some(poll_attempt), &mut mfa_req, None)
+                    .acquire_token_by_mfa_flow(
+                        account_id,
+                        None,
+                        Some(poll_attempt),
+                        &mut mfa_req,
+                        None,
+                    )
                     .await
                 {
                     Ok(token) => break token,
--- a/src/common/src/auth.rs
+++ b/src/common/src/auth.rs
@@ -48,14 +48,12 @@ use tokio::runtime::Runtime;
 macro_rules! auth_handle_mfa_resp {
     ($resp:ident, $on_fido:expr, $on_prompt:expr, $on_poll:expr) => {
         match $resp.get_default_mfa_method_details() {
-            Some(value) => {
-                match value.auth_method_id.as_str() {
-                    "FidoKey" => $on_fido,
-                    "AccessPass" | "PhoneAppOTP" | "OneWaySMS" | "ConsolidatedTelephony" => $on_prompt,
-                    _ => $on_poll,
-                }
-            }
-            None => $on_poll
+            Some(value) => match value.auth_method_id.as_str() {
+                "FidoKey" => $on_fido,
+                "AccessPass" | "PhoneAppOTP" | "OneWaySMS" | "ConsolidatedTelephony" => $on_prompt,
+                _ => $on_poll,
+            },
+            None => $on_poll,
         }
     };
 }
--- a/src/pam/src/pam/mod.rs
+++ b/src/pam/src/pam/mod.rs
@@ -509,7 +509,7 @@ impl PamHooks for PamKanidm {
                     None,
                     &auth_options,
                     Some(auth_init),
-                    cfg.get_mfa_method().as_deref()
+                    cfg.get_mfa_method().as_deref(),
                 )
                 .await
             }) {
@@ -582,8 +582,14 @@ impl PamHooks for PamKanidm {
                         }
                     };
                     match rt.block_on(async {
-                        app.acquire_token_by_mfa_flow(&account_id, Some(&input), None, &mut mfa_req, None)
-                            .await
+                        app.acquire_token_by_mfa_flow(
+                            &account_id,
+                            Some(&input),
+                            None,
+                            &mut mfa_req,
+                            None,
+                        )
+                        .await
                     }) {
                         Ok(token) => token,
                         Err(e) => {
@@ -612,7 +618,7 @@ impl PamHooks for PamKanidm {
                                 None,
                                 Some(poll_attempt),
                                 &mut mfa_req,
-                                None
+                                None,
                             )
                             .await
                         }) {
--- a/src/common/src/config.rs
+++ b/src/common/src/config.rs
@@ -1755,7 +1755,10 @@ mod tests {
         let config = HimmelblauConfig::new(Some(&temp_file)).unwrap();
 
         // Test when MFA method is set
-        assert_eq!(config.get_mfa_method(), Some("TwoWayVoiceMobile".to_string()));
+        assert_eq!(
+            config.get_mfa_method(),
+            Some("TwoWayVoiceMobile".to_string())
+        );
 
         // Test when config is missing (should return None)
         let config_empty = HimmelblauConfig::new(None).unwrap();


Makefile
```diff
@@ -41,6 +43,20 @@ PLATFORM := $(shell grep '^ID=' /etc/os-release | awk -F= '{ print $$2 }' | tr -
...
+  LIBHIMMELBLAU_BUILD_ARG := --build-arg "CARGO_PATCH_ARG=--config 'patch.crates-io.libhimmelblau.path=\"/libhimmelblau\"'"
+  PATCH_LIBHIMMELBLAU := --patch-libhimmelblau
+  $(info Using local libhimmelblau from: $(LIBHIMMELBLAU_LOCAL))
+  $(info LIBHIMMELBLAU_BUILD_ARG: $(LIBHIMMELBLAU_BUILD_ARG))
+else
+  LIBHIMMELBLAU_MOUNT :=
+  LIBHIMMELBLAU_BUILD_ARG :=
+  PATCH_LIBHIMMELBLAU :=
+endif
+
```
dmulder, line 59: Nice! I've been doing something similar (manually) every time I build with a locally modified libhimmelblau. This approach is slick.

man/man5/himmelblau.conf.5
```diff
@@ -185,6 +185,30 @@ Determines whether password-only (single-factor) authentication is permitted whe
...
+.IP \(bu 2
+TwoWayVoiceMobile - Phone call to mobile
+.IP \(bu 2
+TwoWayVoiceOffice - Phone call to office
+.IP \(bu 2
+ConsolidatedTelephony - Call or text message
+.RE
+
+.EXAMPLES
+mfa_method = TwoWayVoiceMobile
```
dmulder, line 191: You are missing "CompanionAppsNotification" (this is MFA via the Outlook app, IIRC) and "TwoWayVoiceAlternateMobile".

scripts/gen_dockerfiles.py
```diff
@@ -314,9 +318,20 @@ def build_rpm_final_cmd(features: list, selinux: bool) -> str:
...
+
 # Install essential build dependencies
 {bootstrap}
 
+# Install Rust (latest stable), verifying sha256sum of rustup-init.sh script
+# current sha256 sum is for rustup-init 1.28.2 (d1f31992a 2025-04-28)
+RUN curl --proto '=https' --tlsv1.2 -sSf -o rustup-init.sh https://sh.rustup.rs && \\
+    echo "17247e4bcacf6027ec2e11c79a72c494c9af69ac8d1abcc1b271fa4375a106c2 rustup-init.sh" | sha256sum --check --status && \\
+    sh ./rustup-init.sh -y && \\
+    /root/.cargo/bin/rustc --version
```
dmulder: What happens here when rustup is updated to a newer version? Do we start to fail at that point?

scripts/gen_dockerfiles.py
```diff
@@ -326,9 +341,10 @@ def build_rpm_final_cmd(features: list, selinux: bool) -> str:
 # Change directory to the repository
 WORKDIR /himmelblau
 
-# Install Rust (latest stable)
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y && \\
-    cargo install cargo-deb cargo-generate-rpm
+# Install the cargo-deb and cargo-generate-rpm tools
+RUN --mount=type=cache,target=/usr/local/cargo/registry cargo install cargo-deb cargo-generate-rpm
```
dmulder: I'm confused by the syntax here. I thought `RUN` was a shell command? Why here does the command start with parameters? Is this some Docker-ism that I'm just not familiar with?

src/common/src/auth.rs
```diff
@@ -47,10 +47,18 @@ use tokio::runtime::Runtime;
...
+            Some(value) => {
+                match value.auth_method_id.as_str() {
+                    "FidoKey" => $on_fido,
+                    "AccessPass" | "PhoneAppOTP" | "OneWaySMS" | "ConsolidatedTelephony" => $on_prompt,
+                    _ => $on_poll,
+                }
+            }
+            // TODO: Will there always be a default MFA method?
+            // If get_default_mfa_method_details() doesn't return a value
+            // we can use, maybe we need to raise an error here?
```
dmulder: MS sometimes doesn't set a default. In that case, we should take the first MFA method suggested. Looking at the implementation of `get_default_mfa_method_details`, there is a chance we might not respond here. Perhaps we should modify that function in libhimmelblau to always return a value (so long as there is ANY method available). See my implementation of the `mfa_method` on the same struct.

src/common/src/idprovider/himmelblau.rs
```diff
@@ -1904,17 +1905,42 @@ impl IdProvider for HimmelblauProvider {
...
+                        .await
+                        .initiate_acquire_token_by_mfa_flow_for_device_enrollment(
+                            account_id,
+                            Some(&cred),
+                            &opts,
+                            None,
+                            None, /* MFA method */
+                        )
+                        .await
+                };
```
dmulder: This should be collapsed into a single call to `initiate_acquire_token_by_mfa_flow_for_device_enrollment`. Just pass `mfa_method` into the last option.

src/common/src/idprovider/himmelblau.rs
```diff
@@ -1307,6 +1307,7 @@ impl IdProvider for HimmelblauProvider {
                                 None,
                                 &auth_options,
                                 Some(auth_init),
+                                None /* MFA method */
```
dmulder: `self.config.read().await.get_mfa_method()` should have been passed here. This is the entry point for passwordless authentication.

src/pam/src/pam/mod.rs
```diff
@@ -488,21 +488,49 @@ impl PamHooks for PamKanidm {
...
+                            None, /* MFA method */
+                        )
+                        .await
+                    }) {
+                        Ok(mfa) => mfa,
+                        Err(e) => {
+                            error!("{:?}", e);
+                            return PamResultCode::PAM_AUTH_ERR;
+                        }
+                    }
```
dmulder: This is a one-line change, if you just pass `cfg.get_mfa_method()` as the last parameter to `initiate_acquire_token_by_mfa_flow`.

src/pam/src/pam/mod.rs
```diff
@@ -488,21 +488,49 @@ impl PamHooks for PamKanidm {
...
+                            &auth_options,
+                            Some(auth_init),
+                            Some(&mfa_method),
+                        )
+                        .await
+                    }) {
+                        Ok(mfa) => mfa,
+                        Err(e) => {
+                            error!("Failed to initiate MFA with configured method '{}': {:?}", mfa_method, e);
+                            return PamResultCode::PAM_AUTH_ERR;
```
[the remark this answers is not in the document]
dmulder: I don't think this is correct.

src/common/src/auth.rs
```diff
@@ -47,10 +47,18 @@ use tokio::runtime::Runtime;
...
+            None => $on_poll
         }
```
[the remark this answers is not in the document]
dmulder: That's the expected behavior.

scripts/gen_dockerfiles.py
surry: Yeah, the RUN command in a Dockerfile supports mounting in paths that will only be available during the image build:

https://docs.docker.com/reference/dockerfile/#run---mount

man/man5/himmelblau.conf.5
```diff
@@ -185,6 +185,30 @@ Determines whether password-only (single-factor) authentication is permitted whe
 .EXAMPLES
 enable_sfa_fallback = true
 
+.TP
+.B mfa_method
+.RE
+Specifies a preferred MFA (Multi-Factor Authentication) method to use during authentication. When set, Himmelblau will attempt to use this specific MFA method instead of the default method configured in the user's Entra ID profile. If not set or if the specified method is not available for the user, the default MFA method will be used.
```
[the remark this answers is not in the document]
dmulder: I think fallback is happening in libhimmelblau. Am I correct?

Makefile
surry: Cool, I will keep this in then!

scripts/gen_dockerfiles.py
dmulder: Ah, ok.

scripts/gen_dockerfiles.py
surry: Right now, yeah - it looks like I could pull a specific tag from https://github.com/rust-lang/rustup and use the rustup-init.sh from there so the hash matches. But then this wouldn't always use the latest version of rustup / rust. I usually try to verify the hashes in Dockerfiles when installing via curl like this to help prevent something malicious from being installed, but this could be saved for another PR.

src/common/src/idprovider/himmelblau.rs
surry: Oops - thanks. Think I had some other differences in these calls when I had my WIP libhimmelblau using the v2 endpoint and didn't realize the redundancy

src/pam/src/pam/mod.rs
surry: I'll also revert the "offline_access" scope I added - that was needed when testing with the v2 endpoint.

scripts/gen_dockerfiles.py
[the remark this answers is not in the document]
surry: Made this change in my next commit I'll push up. the cargo path in the container is /root/.cargo so this seems correct.

src/common/src/auth.rs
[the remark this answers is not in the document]
surry: I kept the existing behavior here for the on_fido/on_prompt - guessing a bit here, but PhoneAppNotification is when you approve an MFA prompt on an authenticator app and you don't need to type in a code? So the default of on_poll here seems OK...

src/common/src/auth.rs
surry: Ok, the 2 copilot suggestions above about falling back to a default also seem related to this. Will think about this some more

scripts/gen_dockerfiles.py
dmulder: Either way, but if we start pulling a specific tag version, we should probably implement a CI job for updating the version we've tagged periodically. In that case, it might be better to push this change to a different PR.

src/common/src/idprovider/himmelblau.rs
dmulder: You marked this resolved, but didn't change the line :confused:

surry: Sorry, think I made this change on https://github.com/himmelblau-idm/himmelblau/pull/800/files#diff-620264a8068ba00ab4d563fdc03c983016e919fabf01d491823ce1968f3941abR1919 (line 1919) below and in pam/mod.rs, but forgot to make it here.

src/common/src/auth.rs
surry: Opened an MR for this in libhimmelblau: https://gitlab.com/samba-team/libhimmelblau/-/merge_requests/124

None can still be returned here, but in this case, we are just determining whether to poll or prompt for the MFA response. If there somehow no valid MFA method, then I would think we'd error out elsewhere (possibly before even getting to this point).

man/man5/himmelblau.conf.5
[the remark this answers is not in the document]
surry: It doesn't look like that's happening in libhimmelblau right now - is that the right place for it, or should it be handled here?

src/common/src/idprovider/himmelblau.rs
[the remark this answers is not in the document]
surry: Similar to the other thread on the man page about what happens if an invalid/unavailable method is specified. Need to figure out if this fallback should happen in himmelblau or libhimmelblau.

scripts/gen_dockerfiles.py
surry: I've reverted this change in my latest commits.

src/cli/src/main.rs
```diff
@@ -250,6 +250,7 @@ async fn auth(app: &BrokerClientApplication, account_id: &str) -> Option<UserTok
             password.as_deref(),
             &auth_options,
             Some(auth_init),
+            None, /* MFA method */
```
surry: Should the CLI pass in the new configuration value here too? It looks like it reads the same config file as the daemon.

src/common/src/idprovider/himmelblau.rs
surry: This should be resolved now.

src/cli/src/main.rs
dmulder: Yes, I think it should.

src/cli/src/main.rs
```diff
@@ -281,6 +282,7 @@ async fn auth(app: &BrokerClientApplication, account_id: &str) -> Option<UserTok
                         password.as_deref(),
                         &auth_options,
                         Some(auth_init),
+                        None, /* MFA method */
```
dmulder: And probably here too.

src/common/src/idprovider/himmelblau.rs
```diff
@@ -2344,7 +2349,7 @@ impl IdProvider for HimmelblauProvider {
                     self.client
                         .read()
                         .await
-                        .acquire_token_by_mfa_flow(account_id, Some(&assertion), None, flow)
+                        .acquire_token_by_mfa_flow(account_id, Some(&assertion), None, flow, None)
```
dmulder, line 2352: It looks like we can (and probably should) remove the auth-method arg from `acquire_token_by_mfa_flow` in libhimmelblau.

dmulder: Because we supplied it to the initiate function, and we've already decided on a method by this point (and it's too late to change our minds even if we wanted to).

dmulder:
A couple more nit-picks, but looking like we're really close. This needs extensive testing now.

src/common/src/idprovider/himmelblau.rs
surry: I'll create another small MR for this. There were some earlier threads that were resolved that hit on the retry behavior if an unsupported MFA method is configured with this new feature. Right now, I don't think this will fall back to the default MFA method if the configured MFA method is not enabled for the user. Should that be handled in libhimmelblau, or in this PR?

surry:
> A couple more nit-picks, but looking like we're really close. This needs extensive testing now.

Cool - FWIW, I built himmeblau from this branch and have it running on a machine that was recently upgraded to Ubuntu 24.04. It is working as expected there - using my pre-configured MFA method for logins and sudo. 

The only thing I noticed so far was that even if I log in with MFA, then immediately lock the screen and try to log in again, I get another MFA request on the 2nd login. That login works as before, but I was hoping there was some caching to avoid another MFA request. I tried increasing `cache_timeout` to 2 hours but that didn't seem to have an effect. Don't think this is related to my changes - a more general behavior I might not be understanding.

dmulder:
> The only thing I noticed so far was that even if I log in with MFA, then immediately lock the screen and try to log in again, I get another MFA request on the 2nd login. That login works as before, but I was hoping there was some caching to avoid another MFA request. I tried increasing `cache_timeout` to 2 hours but that didn't seem to have an effect. Don't think this is related to my changes - a more general behavior I might not be understanding.

That's actually the expected behavior. `cache_timeout` does not effect MFA. Could we improve this? Probably, but it will require some more extensive work in libhimmelblau. Right now Himmelblau **always** requires MFA, unless you enable SFA fallback, and **don't configure MFA at all for a user**.

The reason for the strict enforcement of MFA is because we don't always know where an authentication request is coming from via PAM. If we demand MFA at GDM, and you successfully provide it, and then we have a grace period... then someone accessing that host remotely can now bypass and access your host via the entra password. It's a potential risk that requires very careful consideration.

Consider this scenario: the host is a public facing server (yes, users are using Himmelblau this way). It's accessed via SSH. A legitimate user (user1) accesses the host via their MFA creds. Right now Himmelblau demands this every time. What if we relaxed that requirement? Assume the very next person to authenticate is a hacker who has stolen user1's entra password. They simply have to wait until the grace window opens, and then they have **MFA** access to the host (they'd be piggy backing on the MFA amr from the actual user1 MFA auth). That's why we don't support this, YET.

src/common/src/idprovider/himmelblau.rs
dmulder: Hrm, I suppose let's leave it as is for now.

PR #806, opened 2025-10-21
Fix behaviour change with local_groups and sudo_groups

### LINKED ISSUE #802: Behavior change in the use of sudo in local_groups in himmelblau-2.0.0-alpha
**Description**
With version **2.0.0-alpha** of himmelblau, it is necessary to define a group ID in **sudo_groups** if a user must belong to the sudo group. If sudo_groups is not defined, the user is removed from the sudo group in **/etc/groups**, to which they had been added with **local_groups**. More information in ticket #711.

This change in behavior can be a problem on production systems, where users will lose their sudo capabilities unless sudo_groups is defined.
In large environments, such as universities and labs, computer lab administrators who are not Azure administrators lose the ability to define sudo access.

Ideally, if the sudo group is already defined in local_groups, it should not be deleted unless sudo_groups exists and is defined in himmelblau.conf.

On the other hand, it is a good idea to define a group ID in sudo_groups, ensuring that not all users who authenticate with their Entra account have sudo rights. 

Both behaviors, local_groups and sudo_groups, are acceptable, and I don't have a specific opinion on what should be done. But, at the very least, **this change in behavior should be noted in order to minimize the effect** on production environments, which may even update the version automatically through the repositories.

**Environment**
- **Neon Linux 24.04 / Ubuntu 24.04.3**: 
- **Himmelblau Version**: 2.0.0-alpha

---

### 💡 Help Make It Happen!
Want to see this bug fixed faster? Fund its implementation through our **Backer's Bounty** program, where you choose which bug fixes get priority!

For US tax exempt donations:

iLikeToCode:
Fixes #802

Testing shortly, just gonna let CI get a head start as it builds fine.

### DIFF
--- a/src/selinux/src/himmelblaud.te
+++ b/src/selinux/src/himmelblaud.te
@@ -66,7 +66,8 @@ allow init_t himmelblau_var_lib_t:dir add_name;
 allow init_t himmelblau_var_lib_t:file { create open write };
 accountsd_manage_lib_files(init_t)
 corenet_tcp_connect_http_port(init_t)
-files_manage_home_dir(init_t)
+files_search_home(init_t)
+userdom_manage_user_home_content(init_t)
 
 ############################
 # himmelblaud-tasks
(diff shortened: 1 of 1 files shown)


iLikeToCode:
Same here - it looks good but it doesn't fix the issue

dmulder:
Doh! Sorry! The line needs to have `trim()` first, and then `to_string()`, like this:

```rust
Some(val) => val.split(',').map(|s| s.trim().to_string()).collect(),
```

PR #833, opened 2025-11-03
Add support for Fedora 43

bror-lauritz:
Fedora 43 was released on the 28th of October and should be supported. Builds fine on my machine and functions well on my F43 rig.

Fedora 41 will be deprecated on the 25th of November, but I've not removed it in this PR.


dmulder:
Thank you for contributing!

dmulder:
I think we should additionally pull Fedora 41 from 2.0, if it's being deprecated.

dmulder:
@bror-net FYI: https://packages.himmelblau-idm.org/nightly/2025-11-04-712364ab183e/rpm/fedora43/

PR #868, opened 2025-11-13
Stable backports

dmulder:


dmulder:
This includes some patches to fix the o365 build on the build service host (forcing a rebuild of the .desktop apps on every build).

dmulder:
I've also now included a change in Intune behavior. A failure in Intune compliance used to cause the authentication to be rejected. This is no longer the case, since sometimes the Intune protocol may go down when Entra Id is still active. This could cause authentication failures, when in reality the protocol is working as intended (MS has their clients ignore Intune outages).

PR #884, opened 2025-11-20
Stable osc build fixes

dmulder:
Fixes #


dmulder:
This PR is intentionally skipping the revision change, since it is only an update intended for downstream openSUSE builds.

PR #910, opened 2025-12-02
Add personal account authentication (MSA auth)

dmulder:
This adds behavior that mimics Windows, permitting users to login using their MS enrolled personal email accounts.


dmulder:
Hello PINs are working, but SSO is not yet (and much of it can't work).
For SSO, we're caching a refresh_token instead of a PRT. This allows access token requests to work, but prt cookie requests will fail (and IIUC, there is no working around this).
We'll need to work with SSO providers to ensure they fallback to a access_token request if the prt cookie request fails.

dmulder:
I've fixed the SSO access token issues, but I've also verified that linux-entra-sso (the SSO plugin we use) only utilizes the PRT cookie for SSO (which we **cannot** provide for personal accounts). We'll need to work with upstream to fix this.

PR #918, opened 2025-12-06
systemd: install files to /usr/lib/, not /etc/

bluca:
/etc/ is for local configurations by the users/admins. Packages must install to /usr/lib/, so that users can override them locally if they want to do so.
Packages need to be resilient against factory resets (deleting /etc/ and /var/ and rebooting).

Also add a more descriptive, namespaced name for the gdm3 override, as override.conf is too generic and could conflict.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ rpm-servicefiles:
 	python3 ./scripts/gen_servicefiles.py --out ./platform/opensuse/
 
 authselect:
-	python3 ./scripts/gen_authselect.py --root=./ --aad-tool=./target/debug/aad-tool --output-dir=./platform/el/authselect/
+	python3 ./scripts/gen_authselect.py --root=./ --aad-tool=./target/release/aad-tool --output-dir=./platform/el/authselect/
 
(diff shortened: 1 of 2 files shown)


dmulder:
Would you also update the `cargo-generate-rpm` files to do the same?

bluca:
> Would you also update the `cargo-generate-rpm` files to do the same?

it already does:

```
[package.metadata.generate-rpm]
name = "himmelblau"
maintainer = "David Mulder <dmulder@suse.com>"
assets = [
  { source = "../../src/config/himmelblau.conf.example", dest = "/etc/himmelblau/himmelblau.conf", mode = "644", config = "noreplace" },
  { source = "../../src/config/krb5_himmelblau.conf", dest = "/etc/krb5.conf.d/", mode = "644" },
  { source = "../../src/config/gdm3_service_override.conf", dest = "/usr/lib/systemd/system/display-manager.service.d/override.conf", mode = "644" },
  { source = "target/release/aad-tool", dest = "/usr/bin/", mode = "755" },
  { source = "../../platform/opensuse/himmelblaud-tasks.service", dest = "/usr/lib/systemd/system/", mode = "644" },
  { source = "../../platform/opensuse/himmelblaud.service", dest = "/usr/lib/systemd/system/", mode = "644" },
```

dmulder:
> it already does:

:man_facepalming: Sorry, I didn't look very close.

PR #942, opened 2025-12-29
NixOS: Deflakeify and bring module up to date

Srylax:
Unfortunately the NixOS module lags behind. I moved it into its own module declaration so that non flake users can use it easy without using flake-compat. I have not yet tested the flake.
It is still a heavy wip because the flake does not work again yet.
The non flake modules are tested and already being used at my company.

A big change was moving away from the custom ini generator and removing the option types.
The downside is now that it is not autocompleted and typesafe but the upside is that we do not have to change the module if a new config value gets added which I think is more important because there is currently little maintenance work done on the nix stuff.

Another minor annoyance is since we switched to the normal ini format lists are no longer possible the nix way.

I also added the firefox and chrome extensions by default - maybe make this an additional option?
Furthermore I found a bug where default values are not the same when you are using `--config` vs `/etc/himmelblau/himmelblau.conf`

TODOs:
~- [ ] fix pam ordering to load before early unix~ I'll do this in a new PR
- [x] change the docs and add non flake example
- [x] fix flake
- [x] maybe add the generated o365 desktop files
- [x] desktop split
- [x] remove the systemd config reload hack

In future work we could switch to crane to enable incremental nix builds. This would be beneficial for devShells


$HOME/.config/o365-profiles/onedrive/Network Persistent State
twoolie, line 1: Did you mean to check in all these files?

nix/functions/o365.nix
twoolie, line 1: Seems handy, but it should probably build a meta-package that exports all the desktop entries from here, instead of attaching them to the himmelblau package. himmelblau on the server should not have to carry the desktop entries around.

sources.nix
twoolie, line 1: this file possibly not needed. non-flake consumers can provide their own preferred instance of nixpkgs/rust-overlay

nix/packages/himmelblau.nix
```diff
@@ -0,0 +1,206 @@
...
+      {
+        name = "David Mulder";
+        email = "dmulder@samba.org";
+        github = "dmulder";
+      }
+    ];
+    platforms = lib.platforms.linux;
+  };
+
+  desktopItems = [
```
twoolie: move this somewhere else please

nix/modules/himmelblau.nix
```diff
@@ -0,0 +1,206 @@
...
+          before = [ "accounts-daemon.service" ];
+          wantedBy = [
+            "multi-user.target"
+            "accounts-daemon.service"
+          ];
+          # This is just so that the service restarts on config change
+          # Normall we would do `--config ${configFile}`
+          # Himmelblau has however a bug where if we supply a config file with `--config`,
+          # the default values for fields do not get set.
+          environment.HIMMELBLAU_DUMMY_CONFIG = configFile;
```
twoolie: surely this is not correct...

twoolie:
Hi @Srylax, thanks for looking into this.

I understand not everyone in the nix community is on-board with flakes, but it was important to me that the nix config could be made as self-contained as possible, since @dmulder does not have the bandwidth to work on it, and i can only look it it every couple months.

I am not opposed to breaking it out into smaller pieces in a nix directory, provided @dmulder agrees, but I would want to make sure that it retains the previous level of functionality before adding additional outputs.

> Unfortunately the NixOS module lags behind.

I do try to keep up with upstream changes, but this is not my focus right now. Patches welcome.

> A big change was moving away from the custom ini generator and removing the option types.

Why? the ini generator was there simply to do type conversion so we don't have to use strings for all the values in settings. The option types are there for building documentation.

> The downside is now that it is not autocompleted and typesafe but the upside is that we do not have to change the module if a 
new config value gets added [...]

Not just typesafe, but *typosafe*. if you typo an option name, you get a build-time error instead of an ignored value. This is pretty important for software in your access control stack. I would rather have a module that correctly models 90% of the options and requires periodic patching to add new ones, than a module that allows obviously broken configurations to fail at runtime. 

> [...] which I think is more important because there is currently little maintenance work done on the nix stuff.

I do plan to keep maintaining the nix code this year, and adding new options to the module is at the top of my list.

>Furthermore I found a bug where default values are not the same when you are using --config vs /etc/himmelblau/himmelblau.conf

can you give more info on this?

twoolie (commented):
Some good ideas in here, I'd like to split the update to 25.11 out and keep this MR focused on moving code out of the flake.

nix/functions/o365.nix
Srylax: yeah absolutely, I just didn't have the time yet. maybe `himmelblau` and `himmelblau-desktop`? It also gets really heavy because of the `teams-for-linux` dependency thereforce it should also be an option to disable it in `himmelblau-desktop`

Srylax:
@twoolie Thank you for your reply and the code review!

About the building out the settings vs just a freeform type. I absolutely get both sides of the coin.
My goal however was going as much to the nixos reference as possible. As stated in RFC 42:
https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md#part-2-1

> The second part of this RFC aims to encourage people to write better NixOS modules in terms of quality, maintainability and discoverability by limiting NixOS options representing single settings to a set of most "valuable" options. The general idea of valuable options is that they provide more value (used by people, provide safety) than the trouble they're worth (bloated option listings, maintenance cost). Of course this isn't something we can measure, so it's up to the module author to make a reasonable decision, but some general suggestions are given in the next section. As more such options are deemed valuable they can be added to the module over time as well.

Many NixOS modules have this approch. Probably our closest neighbor: `services.kanidm.serverSettings`

I will change the type to submodule with freeformType. As i see it the only required option is `services.himmelblau.global.domain`?

sources.nix
Srylax: It keeps the flake inputs and non-flake on the same version. Without it consumers would have to always set nixpkgs

Srylax:
Fixes: #977, #949 (in testing)

dmulder:
FYI, #974 caused a branch conflict. Sorry about that, but the changes will need to be merged here. I also just fixed the failing `cargo vet` issues.

twoolie:
> About the building out the settings vs just a freeform type. I absolutely get both sides of the coin. My goal however was going as much to the nixos reference as possible. As stated in RFC 42: https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md#part-2-1

This RFC covers modules to be included in the nixos repo, not third party modules.

> The general idea of valuable options is that they provide more value (used by people, provide safety) than the trouble they're worth (bloated option listings, maintenance cost). 

This is reasonable for NixOS itself, as it should not have to bear the maintenance cost of keeping up-to-date with upstream projects, however we are bundling this module with the project, so the maintenance cost for us to keep options up-to-date is minimal, and the options are listed only for projects that opt-in to our module. 

> Many NixOS modules have this approch. Probably our closest neighbor: `services.kanidm.serverSettings`

kanidm is bundled into NixOS. If we were to submit this module to nixos, then we may consider dropping the options at that time.

Srylax:
@twoolie I agree with you that the RFC applies to modules in the nixos repo. For me the goal was to maybe even upstream the module once it was tested enough. If it were to be upstreamed, I would like to have compatibility between upstream setting syntax and himmelblaus. Nevertheless, I think the maintenance burden is too high for a fully typed module (just look how outdated the current module is #977
). I do not have the bandwidth to maintain a full typed option modules, because this fork is done for "$WORK". 
If you want to go forth and maintain a  fully typed module you are free to do so and not merge this PR. I will always keep this branch up-to-date.

dmulder:
@twoolie and @Srylax, options are now specified using xml files, which generate the rust integration and the man page. Would it make sense to generate the options for nixos also? 
I admit I don't really understand the disagreement here, as i don't use nixos.

Srylax:
@dmulder That could work, yes. Altough it would be wayyyy easier if the docs were in JSON. Is there some specific reason that himmelblau opts to use formats that are consider "older" e.g.  global INI style instead of systemd INI or even better toml!, XML instead of JSON or does it just boild down to personal preference?

dmulder:
> or does it just boild down to personal preference?

It's entirely due to personal preference :satisfied:

dmulder:
@Srylax I pushed a commit to your branch which implements the option generation.
@twoolie is this approach sufficient?
Full disclosure, this commit is LLM generated. I don't understand at all what it's doing :rofl: so please review carefully.

iLikeToCode (changes requested):
> Variable section_title is not used.

https://github.com/himmelblau-idm/himmelblau/pull/942#discussion_r2687112285

> The --config flag is being passed to himmelblaud, but this could potentially cause issues with the configuration handling. According to the PR description, there's a bug where default values are not the same when using --config vs /etc/himmelblau/himmelblau.conf. Since the config is already placed at /etc/himmelblau/himmelblau.conf (line 92), consider removing the --config flag and letting himmelblaud use its default config path, or verify that the --config behavior has been fixed.

https://github.com/himmelblau-idm/himmelblau/pull/942#discussion_r2687112285

Other than that LGTM.

nix/modules/himmelblau.nix
```diff
@@ -0,0 +1,231 @@
...
+          before = [ "accounts-daemon.service" ];
+          wantedBy = [
+            "multi-user.target"
+            "accounts-daemon.service"
+          ];
+
+          upholds = [ "himmelblaud-tasks.service" ];
+          serviceConfig = commonServiceConfig // {
+            ExecStart =
+              "${cfg.package}/bin/himmelblaud --config ${configFile}" + lib.optionalString cfg.debugFlag " -d";
```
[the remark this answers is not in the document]
Srylax: Somewhere we need a reference to the config file in the systemd service to replace it when the config changes. Since we still create the file in /etc it bug does not appear

nix/modules/himmelblau.nix
```diff
@@ -0,0 +1,231 @@
...
+      nativeMessagingHosts.packages = [ cfg.package ];
+    };
+
+    programs.chromium.extensions = [
+      "jlnfnnolkbjieggibinobhkjdfbpcohn"
+    ];
+    environment.etc."chromium/native-messaging-hosts/linux_entra_sso.json".source =
+      "${cfg.package}/lib/chromium/native-messaging-hosts/linux_entra_sso.json";
+    environment.etc."opt/chrome/native-messaging-hosts/linux_entra_sso.json".source =
+      "${cfg.package}/lib/chromium/native-messaging-hosts/linux_entra_sso.json";
```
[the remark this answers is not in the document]
Srylax: they only get added if the module is enabled. also firefox itself is not enabled. This mimics how himmelblau installs the extension on other systems

Srylax:
The roff cleanup does not really work:
```
P
\fBNote:\fR In the Azure Portal for the application corresponding to \fBapp_id\fR, ensure that the redirect URI
\fIhimmelblau://Himmelblau.EntraId.BrokerPlugin\fR
```
```
\(bu 2
PhoneAppNotification
```

I could not get it to work with simple regex parsing of the roff. It would be nice if it would spit out markdown.
We could require the `pre-commit` to require pandoc to be installed to handle the conversion but I have not tested how well that works. Furthermore, I do not like the pandoc requirement.

On the nix side it would maybe also make sense to set default values to non null if the value has a default. This would make it more transparent to the module if a default value changes and would make it possible to reference the values in the systemd file. Example:
instead of 
```nix
unitConfig = {
  ConditionPathExists = "/var/run/himmelblaud/task_sock";
};
```
we could write
```nix
unitConfig = {
  ConditionPathExists = "${cfg.settings.task_socket_path}";
};
```

dmulder:
@Srylax I just pushed a patch (and rebased against main) that resolves the issues you pointed out (I believe).

dmulder:
@twoolie please review.

twoolie:
> options are now specified using xml files, which generate the rust integration and the man page. Would it make sense to generate the options for nixos also?

as long as the script is run often enough with `--gen-nix` to keep the option definitions up to date, i'm happy.

> I admit I don't really understand the disagreement here, as i don't use nixos.

The disagreement is whether to apply the nixos policy of allowing "free-form"/untyped configuration passed through to the service to cut down on maintennance burden, or to provide some type/typo checking and documentation by keeping the options up-to-date.

> @twoolie please review.

I like the idea of a script that generates the nix definitions. 👍

My issue is that it produces a flat option set. in the original flake.nix I separated the options that applied to the whole service (`db_path`, `hsm_pin_path`, `socket_path`, etc.) and options that apply only to a specific domain (`tenant_id`, `local_groups`, etc.). The script should be checking the `section` or `domain_specific` attributes to decide which set to place them in. 

Currently it seems like it just looks for options set to an attrset, which none of the options allow. This would be a step backward since users would no longer be able to override global settings per-domain.

dmulder:
> Currently it seems like it just looks for options set to an attrset, which none of the options allow. This would be a step backward since users would no longer be able to override global settings per-domain.

We don't have per-domain settings anymore. They never really worked, and few people were using them. They were deprecated in 2.x.

twoolie:
Oh, well that's all ok then. I should have looked at the release notes :)

PR #971, opened 2026-01-09
Refactor issue-created.yml to have one donation message

iLikeToCode:


.github/workflows/issue-created.yml
dmulder: Don't you need to change this?

dmulder (changes requested):
LGTM

.github/workflows/issue-created.yml
iLikeToCode: goodness glad someone reviews my changes

PR #976, opened 2026-01-10
Add Stale Workflow

iLikeToCode:


Srylax:
Not a big fan of stalebots https://drewdevault.com/2021/10/26/stalebot.html

dmulder:
Would it make sense to flag an issue as stale, but never actually close an issue? Maybe as a way to poke people invested in an issue to try it again?
Or maybe we should rework this altogether, and instead of flagging issues as stale, we have a bot that pokes the reporter to try it again on a newer version?

iLikeToCode:
> Not a big fan of stalebots https://drewdevault.com/2021/10/26/stalebot.html

I saw that article, and I specifically enabled it only for bugs and not for features :)



> Would it make sense to flag an issue as stale, but never actually close an issue? Maybe as a way to poke people invested in an issue to try it again? 
This bot supports disabling stale.

> Or maybe we should rework this altogether, and instead of flagging issues as stale, we have a bot that pokes the reporter to try it again on a newer version?

We could, but I don't really have time to work on that right now.

Srylax:
A possible way that actually adds value would be to add a version label to each bug request e.g. "Reported in 2.0". And then once a new version e.g. 3.0 is released, the workflow would reping all issues <3.0 to request to test it on the new version.

dmulder:
> A possible way that actually adds value would be to add a version label to each bug request e.g. "Reported in 2.0". And then once a new version e.g. 3.0 is released, the workflow would reping all issues <3.0 to request to test it on the new version.

I like that idea.

PR #1002, opened 2026-01-15
Add OBS automation script and fix RPM spec generation

dmulder:
scripts/obs_update.py:
- New automation script for OBS package updates
- Branches OBS package, runs osc service, regenerates spec
- AI-assisted debugging with Claude/Gemini CLI integration
- Compares original vs fixed spec to help update generator

scripts/gen_rpm_spec.py:
- Add SUSE-specific conditionals for authselect (non-SUSE only)
- Add SELinux conditionals (Tumbleweed/SLE 16+ only)
- Fix package names: systemd-mini, selinux-policy-devel, selinux-tools
- Use %{_mandir} macro with wildcards for man pages
- Mark only /etc files as %config
- Set correct permissions (0644) for .desktop/.png/.css files
- Add BuildArch: noarch to qr-greeter package
- Add ghost file for /var/lib/private/himmelblaud
- Add Suggests: authselect to pam-himmelblau
- Add rc* symlinks, tmpfilesdir, icons directory creation
- Add qr-greeter files section

Fixes #


dmulder:
This patch significantly improves openSUSE (downstream) updates for Himmelblau.

PR #1012, opened 2026-01-19
NixOS: Change bash location to nixos location

iLikeToCode:


iLikeToCode:
Of course that won't work, that's the do not edit file, what am I doing....

PR #1013, opened 2026-01-19
OIDC: remove trailing slash trimming for issuer url

iLikeToCode:
If the issuer has a trailing /, then currently the config strips it. The correct way is to leave the daemon to strip it and allow the issuer url to be valid.

### DIFF
--- a/src/common/src/config.rs
+++ b/src/common/src/config.rs
@@ -812,7 +812,6 @@ impl HimmelblauConfig {
             s.trim()
                 .strip_suffix("/.well-known/openid-configuration")
                 .unwrap_or(s.trim())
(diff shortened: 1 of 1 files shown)


dmulder:
I don't follow. Why is it bad to trim the trailing slash?

iLikeToCode:
Because my issuer contains a trailing / and otherwise it panics, the issuer found in the oidc discovery has to exactly match the one in the config.

iLikeToCode:
The / is still stripped when fetching the openid config, but like I said, the issuer must exactly match the one in that file. Currently a trailing / in the code is ‘fixed’ by assuming it is a user mistake, which is invalid.

PR #1037, opened 2026-01-28
/var/cache/private/himmelblaud should not be created tmpfiles

dmulder:
This is created automatically by systemd at
startup.

Fixes #


dmulder:
This needs some testing. I added this when trying to get SELinux working. This might break SELinux again.

dmulder:
FYI, this was preventing a new install of Himmelblau on Ubuntu 24.04 from starting.

PR #1056, opened 2026-02-02
Prohibit kanidm update for stable-2.x branch

dmulder:
These updates break the SLE16 build.


dmulder:
I mistakenly put these in the stable-2.x branch, which dependabot ignores. It only reads rules from the main branch.

Issue #1078, opened 2026-02-06
pam_himmelblau: split_username error prevents local password updates


min-sm-weng:
### Description

When attempting to change the password for a local user (non-domain user) using sudo passwd <username>, the process fails with an "Authentication token manipulation error". The logs show that pam_himmelblau triggers a split_username error, which seems to block the PAM stack from proceeding to pam_unix.so.

### Steps to Reproduce

Environment:
- Ubuntu 24.04.3 LTS
- pam-himmelblau 2.3.4
- nss-himmelblau 2.3.4
- himmelblau-sshd-config 2.3.4
- sqlite3 3.45.1

Steps:
1. Configure Himmelblau:
Edit the configuration: `sudo vim /etc/himmelblau/himmelblau.conf`

2. Restart Services:
`sudo systemctl restart himmelblaud-tasks.service himmelblaud.service`
 
3. Attempt to change a local account password:
 `sudo passwd aaa`

Workaround:
Disable the Azure Entra ID authentication feature using `sudo pam-auth-update`. 

### Relevant log output

```shell
ubuntu@ip-10-11-0-90:~$ id aaa
uid=5501(aaa) gid=5501(aaa) groups=5501(aaa)
ubuntu@ip-10-11-0-90:~$ sudo passwd aaa
2026-02-06T02:25:29.056647Z ERROR split_username
passwd: Authentication token manipulation error
passwd: password unchanged
```

### Version

2.3.4

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct
---

### 💡 Help Make It Happen!

Want to see this issue resolved faster? Fund its implementation through our **Backer's Bounty** program, where you choose which issues get priority!

[![Donate to Our Collective](https://opencollective.com/himmelblau/donate/button.png?color=blue)](https://himmelblau-idm.org/backers.html#backers-bounty)

For US tax exempt donations:

[![Donate US Tax Exempt](https://www.spi-inc.org/images/paypal_btn_donateCC_LG.gif)](https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=8GYKFLX9UNCH2)

Your support helps drive Himmelblau’s evolution!

PR #1080, opened 2026-02-06
Return PAM_IGNORE in chauthtok for local users

khronosd:
When the target user has no UPN (split_username returns None), treat them as a local account and return PAM_IGNORE so the PAM stack continues to pam_unix. Skip before touching the daemon so local password changes (for example, sudo passwd <local_user>) never depend on himmelblaud.

Fixes: https://github.com/himmelblau-idm/himmelblau/issues/1078

### DIFF
--- a/src/pam/src/pam/mod.rs
+++ b/src/pam/src/pam/mod.rs
@@ -468,23 +468,18 @@ impl PamHooks for PamKanidm {
         // Local user (no UPN): not a Himmelblau/Entra account. Skip before touching the
         // daemon so local password changes (e.g. sudo passwd <local_user>) never depend
         // on himmelblaud and continue to pam_unix.
-        if split_username(&account_id).is_none() {
-            debug!(%account_id, "chauthtok: not a UPN, skipping (local user)");
-            return PamResultCode::PAM_IGNORE;
-        }
+        let (_, domain) = match split_username(&account_id) {
+            Some(resp) => resp,
+            None => {
+                debug!(%account_id, "chauthtok: not a UPN, skipping (local user)");
+                return PamResultCode::PAM_IGNORE;
+            }
+        };
 
         let mut daemon_client = match DaemonClientBlocking::new(cfg.get_socket_path().as_str()) {
             Ok(dc) => dc,
             Err(e) => {
                 error!(err = ?e, "Error DaemonClientBlocking::new()");
-                return PamResultCode::PAM_SERVICE_ERR;
-            }
-        };
-
-        let (_, domain) = match split_username(&account_id) {
-            Some(resp) => resp,
-            None => {
-                error!(%account_id, "chauthtok: split_username failed unexpectedly");
                 return PamResultCode::PAM_SERVICE_ERR;
             }
         };


dmulder:
Hrm, I think this will break for OIDC users who have no upn.

dmulder:
> Hrm, I think this will break for OIDC users who have no upn.

Then again, OIDC users can't use this flow to change their passwords anyway.

khronosd:
@dmulder To clarify, are you thinking that if you go to sign in as just "khronosd" instead of "khronosd@contos.com" would fail?

If so, and like you mentioned, they wouldn't go through this anyway and I don't see a regression otherwise.

src/pam/src/pam/mod.rs
```diff
@@ -473,13 +481,7 @@ impl PamHooks for PamKanidm {
...
         };
 
-        let (_, domain) = match split_username(&account_id) {
-            Some(resp) => resp,
-            None => {
-                error!("split_username");
-                return PamResultCode::PAM_AUTH_ERR;
-            }
-        };
+        let (_, domain) = split_username(&account_id).expect("UPN already validated above");
```
dmulder: We should never use `expect` please.

src/pam/src/pam/mod.rs
```diff
@@ -465,6 +465,14 @@ impl PamHooks for PamKanidm {
             None => cfg.map_name_to_upn(&account_id),
         };
 
+        // Local user (no UPN): not a Himmelblau/Entra account. Skip before touching the
+        // daemon so local password changes (e.g. sudo passwd <local_user>) never depend
+        // on himmelblaud and continue to pam_unix.
+        if split_username(&account_id).is_none() {
```
dmulder: Perhaps we could rework this so we only ever call `split_username` once?

dmulder (changes requested):

src/pam/src/pam/mod.rs
khronosd: Good point and will note that down.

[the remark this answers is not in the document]
khronosd: https://github.com/himmelblau-idm/himmelblau/pull/1080/commits/7e03b8f3976224f4ecdf2c453a042eadd7547495

khronosd: https://github.com/himmelblau-idm/himmelblau/pull/1080/commits/7e03b8f3976224f4ecdf2c453a042eadd7547495

src/pam/src/pam/mod.rs
khronosd: https://github.com/himmelblau-idm/himmelblau/pull/1080/commits/3480d3da1f22cee733099e6ff4f14dcef607a6e1

src/pam/src/pam/mod.rs
```diff
@@ -465,6 +465,14 @@ impl PamHooks for PamKanidm {
...
+            debug!(%account_id, "chauthtok: not a UPN, skipping (local user)");
+            return PamResultCode::PAM_IGNORE;
```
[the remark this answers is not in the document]
khronosd: isn't this just overkill given other tests?

src/pam/src/pam/mod.rs
```diff
@@ -465,6 +465,14 @@ impl PamHooks for PamKanidm {
...
+        }
```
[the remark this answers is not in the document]
khronosd: This seems like quite the edge case here and would fall on mapping failing? Otherwise we end up with two checks.

PR #1088, opened 2026-02-08
Fix console password-only MFA bypassing ForceMFA

### LINKED ISSUE #1062: SFA to MFA upgrade fails
### Description

References: https://matrix.to/#/!KvfMtTRFhjDsCHosfQ:matrix.org/$RgnTS0_-9ewLY0Yl9YsFOTkfaM_dASyck8MOFDIluBg?via=matrix.org&via=integrations.ems.host&via=zendis.de

Fix on `dmulder/revert_password_only_login` works.

https://github.com/himmelblau-idm/himmelblau/issues/1045#issuecomment-3820181175

### Steps to Reproduce

1. Create an organization internal network that requires only SFA to login.
2. Enable `allow_console_password_only` and disable hello
3. Login in internal network works
4. Connect to a public network - conditional access policy now requires MFA
5. Login fails
6. You can disable wifi and login using offline auth
7. Now entra_sso plugin does not work because it cannot acquire a token

### Relevant log output

```shell
see matrix
```

### Version

master

### Code of Conduct

---

### 💡 Help Make It Happen!

Want to see this issue resolved faster? Fund its implementation through our **Backer's Bounty** program, where you choose which issues get priority!

For US tax exempt donations:

khronosd:
When `allow_console_password_only` is `true` and the PRT sign-in frequency check fails, the MFA flow was initiated with a cached `auth_init` (fetched without `amr_values=ngcmfa`) and without `ForceMFA` in `auth_options`. This caused Entra to return a password-only session, producing a PRT without MFA claims that fails when exchanging for a Graph access token.

Fix by adding `ForceMFA` to `auth_options` before the first MFA initiation attempt and passing `None` for `auth_init` to force a fresh `request_auth_config_internal()` call with `force_mfa=true`. Also add an explicit `MsalError::MFARequired` handler in `MFAPoll` as a safety net for sessions that lack MFA claims.

Fixes #1062

### DIFF
--- a/.github/workflows/opencode.yml
+++ b/.github/workflows/opencode.yml
@@ -32,6 +32,8 @@ jobs:
           AZURE_COGNITIVE_SERVICES_RESOURCE_NAME: ${{ secrets.AZURE_COGNITIVE_SERVICES_RESOURCE_NAME }}
           AZURE_COGNITIVE_SERVICES_API_KEY: ${{ secrets.AZURE_COGNITIVE_SERVICES_API_KEY }}
           AZURE_API_KEY: ${{ secrets.AZURE_API_KEY }}
+          GH_TOKEN: ${{ github.token }}
+          GITHUB_TOKEN: ${{ github.token }}
         with:
           model: azure-cognitive-services/gpt-5.2-codex
           prompt: |
@@ -87,9 +89,10 @@ jobs:
             const sinceDate = new Date(since);
             const { owner, repo } = context.repo;
             const MAX_ISSUES = 30;
-            const MAX_COMMENTS = 2;
-            const MAX_COMMENT_FETCH = 10;
-            const COMMENT_EXCERPT_CHARS = 200;
+            const COMMENT_BODY_CHARS = 2000;
+            const ISSUE_BODY_CHARS = 2000;
+            const MAX_COMMENTS_PER_ISSUE = 50;
+            const COMMENT_PAGE_SIZE = 100;
             const CONCURRENCY = 5;
             const isHimmelbotAuthor = (author) => {
               const normalized = (author || "").toLowerCase();
@@ -118,11 +121,14 @@ jobs:
                 created_at: issue.created_at,
                 updated_at: issue.updated_at,
                 labels: issue.labels?.map((label) => label.name) || [],
+                body: issue.body ? issue.body.slice(0, ISSUE_BODY_CHARS) : "",
+                body_truncated: Boolean(issue.body && issue.body.length > ISSUE_BODY_CHARS),
                 comment_count: issue.comments || 0,
-                latest_comments: [],
+                comments: [],
                 latest_comment_is_bot: false,
                 bot_commented: false,
                 bot_commented_last_3_days: false,
+                comments_fetch_failed: false,
               };
 
               if (!issue.comments) {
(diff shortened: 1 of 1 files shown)


src/common/src/idprovider/himmelblau.rs
Srylax: unused variable: `auth_init`

Srylax (changes requested):
auth_init: _

src/common/src/idprovider/himmelblau.rs
```diff
@@ -2771,13 +2783,14 @@ impl IdProvider for HimmelblauProvider {
...
-                            Some(auth_init.clone()),
+                            None,
                             self.config.read().await.get_mfa_method().as_deref()
                         )
                         .await,
                     Ok(flow) => flow,
                     Err(MsalError::MFARequired) => {
-                        auth_options.push(AuthOption::ForceMFA);
+                        // If somehow we still get MFARequired, retry
+                        // with ForceMFA (already in auth_options) and no cached auth_init.
```
dmulder: This comment doesn't make sense, given that we already tried with a new auth_init. The retry here might still be necessary (shrug), but this comment doesn't make sense anymore.

khronosd: https://github.com/himmelblau-idm/himmelblau/commit/859ea6fb695bd97de3f1933412235d7085d42ff5

src/common/src/idprovider/himmelblau.rs
```diff
@@ -2761,8 +2759,22 @@ impl IdProvider for HimmelblauProvider {
...
-                debug!("Initiating MFA auth flow");
-
+                // Since password was validated and MFA is required, we must force
+                // interactive MFA to get a session with MFA claims. Without ForceMFA,
+                // Azure may return an auth code directly (password-only session) which
+                // will fail later when exchanging the PRT for a Graph access token.
+                if !auth_options.contains(&AuthOption::ForceMFA) {
+                    auth_options.push(AuthOption::ForceMFA);
+                }
+                debug!("Initiating MFA auth flow with ForceMFA (sign-in frequency not satisfied)");
```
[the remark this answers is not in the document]
dmulder: I think this is correct. We should probably only add ForceMFA if we know we got an MFA demand.

src/common/src/idprovider/himmelblau.rs
```diff
@@ -3114,6 +3126,16 @@ impl IdProvider for HimmelblauProvider {
                                 AuthCacheAction::None,
                             ));
                         }
+                        MsalError::MFARequired => {
+                            error!("MFA required but session lacks MFA claims. \
+                                    This may indicate the MFA flow was initiated without ForceMFA.");
```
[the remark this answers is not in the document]
dmulder: I'm fairly certain this is incorrect.

src/common/src/idprovider/himmelblau.rs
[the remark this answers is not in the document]
khronosd: Agreed. Will update.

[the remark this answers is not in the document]
khronosd: https://github.com/himmelblau-idm/himmelblau/pull/1088/commits/fcc6a61e33d3df72d516d7aca567af6715b44ba7

src/common/src/idprovider/himmelblau.rs
```diff
@@ -2771,13 +2801,14 @@ impl IdProvider for HimmelblauProvider {
...
                             auth_options,
-                            Some(auth_init.clone()),
+                            None,
                             self.config.read().await.get_mfa_method().as_deref()
                         )
                         .await,
                     Ok(flow) => flow,
                     Err(MsalError::MFARequired) => {
-                        auth_options.push(AuthOption::ForceMFA);
+                        // MFA flow initiation returned MFARequired; retry once
```
dmulder: Recall that we might get to this point and _not_ have `ForceMFA` set, so we should add it if missing (in fact, we'll probably only get to this point if it _isn't_ set).

dmulder: So here we'd want to add:
```
if !auth_options.contains(&AuthOption::ForceMFA) {
    auth_options.push(AuthOption::ForceMFA);
}
```

src/common/src/idprovider/himmelblau.rs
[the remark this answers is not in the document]
khronosd: Agreed. Everything is consumed as intended.

src/common/src/idprovider/himmelblau.rs
dmulder: I'm adding this right now.

khronosd: Ah okay. I see this now. I think when I first made these changes, I was still handling it as ForceMFA being added vs conditionally.

PR #1112, opened 2026-02-11
Forbid Hello auth via remote session

dmulder:
This is a _breaking_ change to anyone using
Hello PIN authentication via SSH (an unwise thing
to do, and has been highly discouraged). Hello
via remote (ssh) now requires the Hello TOTP be
enabled as well.

Fixes #


dmulder:
We should make it so this behavior can be disabled via configuration.

PR #1122, opened 2026-02-12
Remove backports script target filtering

dmulder:
Fixes #


dmulder:
Need to allow the user running the script to decline patches for a version with explanation.

PR #1152, opened 2026-02-19
debian: small improvements to package install/remove UX, add separate broker package and unit

bluca:


src/daemon/scripts/postinst
```diff
@@ -19,19 +19,6 @@ systemd-tmpfiles --create /usr/lib/tmpfiles.d/himmelblau-policies.conf 2>/dev/nu
...
-for OLD_FILE in \
-    "/etc/systemd/system/himmelblaud.service" \
-    "/etc/systemd/system/himmelblaud-tasks.service" \
-    "/etc/systemd/system/gdm3.service.d/override.conf"; do
-    if [ -f "$OLD_FILE" ]; then
-        echo "Removing old service file: $OLD_FILE"
-        rm -f "$OLD_FILE"
-    fi
-done
-
```
dmulder, line 34: While you make a good point that users may have defined these intentionally, this hasn't been here very long, because we were putting the service files in the wrong location, and now some users are unable to upgrade due to that prior mistake.

dmulder: The question is, is there any way that we could detect that _we_ created that service file? Probably not, I'd guess, given that people would normally just copy the service file and modify it for their own needs.

bluca: take a hashsum of the version that was installed and compare it, I guess

dmulder: That's not a bad idea. I could go through all the old versions and hashsum them, then detect which ones were something we installed.

src/sso/Cargo.toml
```diff
@@ -25,6 +25,7 @@ broker-client.workspace = true
 name = "himmelblau-sso"
 depends = ["$auto"]
 recommends = ["jq"]
+conflicts = ["linux-entra-sso"]
```
dmulder, line 28: The RPM package sets linux-entra-sso as `provides`.

platform/debian/scripts/postrm
```diff
@@ -49,3 +47,5 @@ if [[ -f /etc/apparmor.d/usr.lib.libreoffice.program.soffice.bin ]]; then
         apparmor_parser -r /etc/apparmor.d/usr.lib.libreoffice.program.soffice.bin >/dev/null 2>&1 || true
     fi
 fi
+
+DEBIAN_FRONTEND=noninteractive pam-auth-update
```
dmulder, line 51: What does this do, exactly?

dmulder: (I'm just ignorant of how Debian installs work)

bluca: It updates the generated pam configs from the definitions. The postinst is already calling this. The package installs definitions, so this is needed to update the generated config after installing/removing. As the commit says, otherwise the pam config still references a non-existing library and falls apart.

src/sso/Cargo.toml
bluca: In deb a provide is not enough, a conflict is needed too. But I can add that too.

src/daemon/scripts/postinst
bluca: I have implemented this with an initial list of hashes, from when the files where checked in the repository. Can I leave it to you to find the hashes of the generated files from shipped versions? The repository only has the latest package version, so cannot check from there

bluca:
I've added a new commit to split the broker binary and dbus config in a new package. As the commit message says, this makes the experience a lot smoother, especially when mixing with linux-entra-sso and the msft broker, as it matches the structure of those packages, with appropriate conflict/replace/provide to make install/remove/upgrade smooth.

I want to try using the browser extension and connector from this repository, but with the msft broker, and this is currently not possible as the sso package contains the broker which conflicts with the msft one (same dbus file is installed).

bluca:
Pushed more changes to further improve the UX. Pretty happy with how the deb package flow looks like now, install/remove/upgrade paths are much smoother, even when mixing with the msft packages and the siemens packages.

As the commit message mentions, the main changes are a new package + user unit for the broker.

src/sso/Cargo.toml
```diff
@@ -23,16 +23,17 @@ broker-client.workspace = true
 
 [package.metadata.deb]
 name = "himmelblau-sso"
-depends = ["$auto"]
+depends = ["$auto", "himmelblau-broker | microsoft-identity-broker"]
```
dmulder, line 26: I'm ok with this, but just a note: Installing Himmelblau and microsoft-identity-broker side-by-side means that you'll have 2 different joined devices and 2 different Intune enrolled devices listed for _every_ host. It would be even better if we could find a way to allow the MS Intune for Linux app to fetch credentials from the Himmelblau Broker, so that they could live more in harmony with one another.

dmulder: What I'd love to see is the option for users to authenticate via Himmelblau, and have MS Intune for Linux enforce policy (working together via a single Entra joined host, etc).

dmulder: FYI, I don't mean that any of that needs to happen in this PR, I'm just voicing my thoughts.

dmulder (approved):
LGTM

src/sso/Cargo.toml
bluca: > I'm ok with this, but just a note: Installing Himmelblau and microsoft-identity-broker side-by-side means that you'll have 2 different joined devices and 2 different Intune enrolled devices listed for every host.

Yes, this is useful when testing and transitioning from one to the other, especially as the CA wasn't working, so I could switch back and forth without completely breaking my ability to do my actual day job :-)

> It would be even better if we could find a way to allow the MS Intune for Linux app to fetch credentials from the Himmelblau Broker, so that they could live more in harmony with one another.
> What I'd love to see is the option for users to authenticate via Himmelblau, and have MS Intune for Linux enforce policy (working together via a single Entra joined host, etc).

I am in no way related to that team, just fixing stuff on my own here, in my own time and initiative and with no connetion whatsoever. From what I can gather, there are basically zero chances of that unfortunately... I agree it would be nice, but nothing I can do about it sadly

PR #1154, opened 2026-02-22
broker: return passwordExpiry too

bluca:
linux-entra-sso and sso-mib expect a passwordExpiry field in the getAccounts response, and fail without it, so include it. With this, they work out of the box with the himmelblau broker.

https://github.com/siemens/linux-entra-sso
https://github.com/siemens/sso-mib


dmulder:
At all related to https://github.com/himmelblau-idm/himmelblau/issues/1159 perhaps?

bluca:
Looks like the same error I had, yes. With this change, linux-entra-sso 1.8.0 works for me.

PR #1156, opened 2026-02-22
broker: add option to allow returning any unlocked and authenticated user

bluca:
Currently only the user logged in via PAM is returned by getAccounts and similar interfaces. Add an option that, when no uid match is found, returns all authenticated and unlocked accounts. This is useful to run or test on existing systems that are not provisioned via Entra, but that use existing local accounts, with different usernames.
The option is disabled by default to preserve existing behaviour.

### DIFF
--- a/docs-xml/himmelblauconf/base/local_account_sso.xml
+++ b/docs-xml/himmelblauconf/base/local_account_sso.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<parameter name="local_account_sso"
+           section="global"
+           type="bool"
+           rust_type="bool"
+           documented="true"
+           domain_specific="false"
+           order="48">
+<description>
+A boolean option that enables browser SSO for local Linux users who did not
+log in via Himmelblau PAM. When enabled, the broker will return cached Entra
+accounts to any local user requesting them, allowing browser extensions such
+as linux-entra-sso to acquire tokens and PRT SSO cookies on behalf of
+enrolled Entra accounts.
+
+.PP
+When disabled (the default), the broker only returns accounts whose UID
+matches the calling process, which requires the user to have logged in
+via Himmelblau PAM/NSS.
+
+.PP
+.B Prerequisite:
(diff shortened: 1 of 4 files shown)


bluca:
An alternative could be to add an option to provide a mapping, eg: "localuser -> entrauser"

dmulder:
I feel like I have to nix this request. When would it ever be appropriate to return the credentials of one signed in user to a different signed in user? You mention this is helpful in testing, but it opens up potential security vulnerabilities.

dmulder:
> An alternative could be to add an option to provide a mapping, eg: "localuser -> entrauser"

We already have local user mapping. See the himmelblau.conf option `user_map_file` for the location of the user map, and the formatting of the map file.

bluca:
> I feel like I have to nix this request. When would it ever be appropriate to return the credentials of one signed in user to a different signed in user? You mention this is helpful in testing, but it opens up potentially security vulnerabilities.

As mentioned in the description, it's because I don't login with the entra user, but with a normal local user. This works with the proprietary implementation and is very useful to access resources behind Conditional Access, without having to fully buy in into entra.

bluca:
> > An alternative could be to add an option to provide a mapping, eg: "localuser -> entrauser"
> 
> We already have local user mapping. See the himmelblau.conf option `user_map_file` for the location of the user map, and the formatting of the map file.

Ah sorry missed that - let me try how it works with the SSO stuff

bluca:
Ok the most important bit works, accessing resources - however it is a bit glitchy, and for example the linux-entra-sso extension doesn't show the account:

image

Is there anything we can do to improve the UI/UX here without compromising security?

dmulder:
> Ok the most important bit works, accessing resources - however it is a bit glitchy, and for example the linux-entra-sso extension doesn't show the account:
> image
> 
> Is there anything we can do to improve the UI/UX here without compromising security?

SSO is supposed to work for mapped local users, so if it isn't working, than this is a bug. So yes, we can look into resolving that. The point of the mapped users feature was to imitate the behavior of MS Intune for Linux.

PR #1259, opened 2026-03-17
feat: auto-sudo for device owner (first logged-in user)

bash-bandicoot:
## Summary
- Adds `auto_sudo_owner` config option (default: false). When enabled,
  the first Entra ID user to log into a device is automatically added
  to the local sudo group (e.g., `sudo` on Ubuntu, `wheel` on RHEL).
  The user still authenticates normally for sudo — this just grants
  group membership automatically.
- The owner is persisted per-domain. Subsequent users are NOT added.
- Adds `aad-tool owner-reset` CLI command to reset ownership (next login
  becomes the new owner). Supports `--domain` to target a specific domain.
- Adds `aad-tool owner-show` CLI command to display the current device
  owner per-domain or for all domains.
- Includes config XML, man page, Nix options, and example config updates.

## Files changed
- `docs-xml/himmelblauconf/base/auto_sudo_owner.xml` — new config option
- `src/daemon/src/daemon.rs` — owner tracking on first login
- `src/cli/src/main.rs` + `opt/tool.rs` — `owner-reset` + `owner-show` subcommands
- `man/`, `nix/`, `src/config/himmelblau.conf.example` — docs/config updates

## Test plan
- [ ] Enable `auto_sudo_owner = true`, log in as first user, verify
      user added to sudo group (`groups <user>`)
- [ ] Log in as second user — verify they are NOT added to sudo group
- [ ] Run `aad-tool owner-show` — displays owner per domain
- [ ] Run `aad-tool owner-show --domain example.com` — shows specific domain
- [ ] Run `aad-tool owner-reset` — next login becomes new owner
- [ ] Run `aad-tool owner-reset --domain example.com` — domain-specific reset
- [ ] Verify `auto_sudo_owner = false` (default) does not add to sudo group

### DIFF
--- a/src/cli/src/main.rs
+++ b/src/cli/src/main.rs
@@ -628,6 +628,10 @@ async fn main() -> ExitCode {
             debug,
             domain: _,
         } => debug,
+        HimmelblauUnixOpt::OwnerShow {
+            debug,
+            domain: _,
+        } => debug,
         HimmelblauUnixOpt::ConfigurePam {
             debug,
             really: _,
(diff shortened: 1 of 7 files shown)


bash-bandicoot:
@dmulder Please review.

src/cli/src/main.rs
```diff
@@ -1664,6 +1672,102 @@ async fn main() -> ExitCode {
...
+                            }
+                        }
+                    }
+                    if !found {
+                        println!("No device owners set.");
+                    }
+                }
+            }
+            ExitCode::SUCCESS
+        }
```
dmulder, line 1770: What is this? Device ownership is a relationship in Entra ID, not some static file placed on the system.

dmulder (changes requested):
This is implemented wrong. If you truly want to implement device ownership changes, you need to perform a relationship change on the device object in Entra ID. There is probably a graph API call you could make that will do this.

src/cli/src/main.rs
bash-bandicoot: Thanks for the feedback, David.
Modifying `registeredOwners` on a device requires `Device.ReadWrite.All`, an admin-level permission that regular users won't have on most tenants.
Calling it with a standard user token would return 403 for the majority of deployments.

The core use case here is granting local sudo on the device itself, which is inherently a local operation. Local tracking also works offline (no network dependency at login time), and survives cache clears. I use atomic file creation (create_new) to handle the race condition safely.

If there's a way to update `registeredOwners` using device credentials rather than a user token, I'd be happy to add that as an enhancement.
Alternatively, if you have a specific Graph endpoint in mind that doesn't require elevated permissions, please let me know, and I'll implement that.

dmulder:
I'm nixing this one. I recommend you implement the behavior you're looking for via a logon_script.

Issue #1234, opened 2026-03-10
System hangs on login interface when an incorrect domain is entered


min-sm-weng:
### Description

The system freezes when an incorrect domain is entered on the first attempt. Even after interrupting the process and providing the correct domain, the login functionality remains broken and fails to authenticate.

In version 2.3.6, the system handled this correctly by providing an immediate error message:
`NotFound: domain: XXX.co in providers`
after interrupting the failed attempt, subsequent logins with the correct domain would still work as expected.

### Steps to Reproduce

Environment:

- Ubuntu 24.04.3 LTS
- himmelblau 3.0.1
- pam-himmelblau 3.0.1
- nss-himmelblau 3.0.1
- himmelblau-sshd-config 3.0.1
- sqlite3 3.45.1

Steps:
1. Configure Himmelblau:
Edit the configuration: `sudo vim /etc/himmelblau/himmelblau.conf`

```
[global]
# These default options are required on Debian/Ubuntu. See `man himmelblau.conf`
# for an explanation of these parameters. `home_attr` and `home_alias` need to
# match, but don't necessarily need to be set to `CN`.
home_attr = CN
home_alias = CN
use_etc_skel = true

domain = XXX.com
pam_allow_groups = XXX
local_groups = docker
enable_hello = true
allow_remote_hello = true
debug = true

[XXX.com]
app_id =XXX
```

2. Restart Services
3. Use incorrect domain login server
```
ssh XXX@XXX.co@10.11.0.74
PAM_IGNORE: An unexpected error occured.
If you are now prompted for a password from pam_unix, please disregard the prompt, go back and try again.
(XXX@XXX.co@10.11.0.74) Password:
```

4. Use correct domain login server
```
ssh XXX@XXX.com@10.11.0.74
PAM_IGNORE: An unexpected error occured.
If you are now prompted for a password from pam_unix, please disregard the prompt, go back and try again.
(XXX@XXX.com@10.11.0.74) Password:
```

### Relevant log output

```shell
Mar 10 06:17:31 ip-10-11-0-74 himmelblaud[10130]: 00000000-0000-0000-0000-000000000000 DEBUG 🐛 [debug]: Sending Task -> "HomeDirectory(...)"
Mar 10 06:17:31 ip-10-11-0-74 himmelblaud[10130]: 00000000-0000-0000-0000-000000000000 DEBUG 🐛 [debug]: Attempting to send request -> "HomeDirectory(...)"...
[...]

Issue #882, opened 2025-11-18
NSS issues when logging in with some users


surry:
**Description**

I'm hitting an issue where logging into a machine graphically with himmelblau is getting the machine into a weird state. I suspect this is not himmelblau itself, but some other configuration on the machine. It previously had `sssd` installed for logging in with LDAP. I have uninstalled sssd and have been testing himmelblau.

I was able to log into the machine with my Entra credentials/MFA, but when I have another user log in for the first time, the UI hangs after entering their username and they don't get prompted for a password.

With a root shell open on the machine remotely, it seems that something with NSS is going wrong. After the new user tries to log in, something like `ls -lah /home/himmelblau-user1` hangs for a long time. It eventually completes (~3 minutes). I ran `ls -lah /home/...` with strace and it seems to be getting blocked trying to read from a domain socket, with a mention of `NssAccountByUid`. It looks like the related code is in here:

https://github.com/himmelblau-idm/himmelblau/blob/main/src/nss/src/implementation.rs#L119

If I restart himmelblaud it seems to get it back into a better state where at least I can log in via himmelblau with my existing account  that was on the machine. A local admin account on the machine also works.

I saw some issues that sounded similar on kanidm, but not really sure how to debug this further.
 
**Logs and Output**

I will try to redact and get relevant logs for this, don't have them at the moment. The only thing in the himmelblau logs that seemed relevant was a timeout on the /var/run himmelblaud socket. 

**Configuration**
```
[global]
domains = mydomain.com
local_groups = users
home_attr = CN
home_alias = CN
use_etc_skel = true
enable_hello = false

pam_allow_groups = ...

sudo_groups = ...

debug = true

mfa_method = TwoWayVoiceOffice

logon_token_app_id = ...
app_id = ...

[offline_breakglass]
enabled = true
ttl = 4h

```

**Environment**
- **OS: ** Ubuntu 24.04: 
- **Source**: Self-built from main
- **Version:** 3.0

dmulder:
I'll need to see debug. Probably include system debug from the journal as well. I recently saw that GDM was crashing for new users, but not for existing users. Perhaps that's what you're seeing here?
I noticed that if I authenticated a new user via some other means (such as with `su`), that the second user was then working.

https://bugs.launchpad.net/ubuntu/+source/gnome-shell/+bug/2130915
https://bugs.launchpad.net/ubuntu/+source/gnome-shell/+bug/2125720

dmulder:
[...]

PR #11, opened 2023-07-10
daemon: Run himmelblaud as non-root dynamic user

dmulder:
This mostly works, but the daemon hangs when trying to create the PublicClientApplication (or maybe it hangs in the auth attempt, I'm not sure). The pam module times out waiting for the daemon to respond, but the daemon does eventually respond with a successful authentication.

PR #1286, opened 2026-04-01
Replace RwLock with Mutex in provider layer to fix deadlocks

moritzheiber:
himmelblaud uses `tokio::main(flavor = "current_thread")` with `tokio::sync::RwLock` on `config`, `client`, `idmap`, `providers`, and `init` in the provider layer. Multiple code paths hold `.read()` guards across `.await` points (HTTP requests to login.microsoftonline.com, odc.officeapps.live.com, etc.), while other paths need `.write()` on the same locks. In the single-threaded runtime this deadlocks permanently:

- Tasks B/C/D hold `config.read()` across HTTP `.await`s
- Task A (e.g. `delayed_init`) needs `config.write()`, which waits for all readers
- Those readers are blocked on `hsm` or `db` locks that Task A holds
- Circular wait, daemon never responds

Using `strace`, the signature is: 2 HTTPS connections succeed (Graph discovery), then the daemon idles in `epoll_wait` forever. The client socket never gets a response.

Similar problems were described in some other issues:

- #882 are somewhat identical symptoms, the reporter confirmed a deadlock via GDB, closed with a workaround that avoided the triggering code path, with the summary being "I suspect there's still a deadlock somewhere."
- #1234 The `get_primary_domain_from_alias()` path calls `request_federation_provider()` (HTTP) while the caller holds `config.write()` via `find_provider!`. Same root cause.
- PR #11 from 2023, "the daemon hangs when trying to create the PublicClientApplication..."

In order to address the underlying issue, I'm replacing `RwLock` with `Mutex` for `config`, `client`, `idmap`, and `providers` in `HimmelblauMultiProvider` / `HimmelblauProvider`, and `init: RwLock<bool>` with `AtomicBool`. `RwLock` allows N concurrent readers, and a single writer must wait for all N to release. `Mutex` allows only one holder at a time, which eliminates the reader-pileup deadlock. Performance impact is negligible since the runtime is single-threaded anyway, there's no real parallelism to lose.

The `init` flag is a simple bool that doesn't need async locking at all, so `AtomicBool` is a good fit.

The broader pattern of holding any lock across HTTP `.await` points is still not ideal, but that's not something you can fix easily with a single PR, so I'm just addressing the deadlock with this one.

Locally, I could reliably reproduce this by running `aad-tool auth-test --name user` (with user mapping) against a freshly started `himmelblaud` where `tenant_id` / `authority_host` are not yet cached. The `delayed_init()` path makes Graph discovery HTTP calls and deadlocks at `client.write()`.

## Required Before Review

- [ ] I manually tested this change on a test VM and confirmed it works as intended.
- [ ] I listed exactly what testing I performed (commands/steps and observed results).
- [ ] I listed which distro(s) and version(s) I tested on.
- [ ] I ran relevant automated checks (for example `make test`, distro package build target, and `make test-selinux` when applicable) or explained why a check was not run.
- [ ] If this change affects system integration (systemd, PAM/NSS/authselect, SELinux/AppArmor, filesystem paths, or credentials), I documented the packaging/runtime impact and any generator/template sources updated.
- [ ] I linked the related issue/enhancement/discussion and confirmed the PR scope is focused and reviewable.

### DIFF
--- a/src/common/src/idprovider/common.rs
+++ b/src/common/src/idprovider/common.rs
@@ -240,7 +240,7 @@ macro_rules! handle_hello_bad_pin_count {
             .increment_bad_pin_count($account_id)
             .await;
 
-        let hello_pin_retry_count = $self.config.read().await.get_hello_pin_retry_count();
+        let hello_pin_retry_count = $self.config.lock().await.get_hello_pin_retry_count();
         let bad_pin_count = $self.bad_pin_counter.bad_pin_count($account_id).await;
 
         if bad_pin_count == hello_pin_retry_count {
@@ -297,7 +297,7 @@ macro_rules! impl_offline_break_glass {
     ($self:ident, $ttl:ident) => {{
         let mut state = $self.state.lock().await;
         let (ttl, enabled) = {
-            let cfg = $self.config.read().await;
+            let cfg = $self.config.lock().await;
             (
                 match $ttl {
                     Some(ttl) => ttl,
@@ -368,7 +368,7 @@ macro_rules! load_cached_prt {
         if let Ok(Some(hello_prt)) = $keystore.get_tagged_hsm_key(&hello_prt_tag) {
             let prt = $self
                 .client
-                .read()
+                .lock()
                 .await
                 .unseal_user_prt_with_hello_key(&hello_prt, &$hello_key, &$cred, $tpm, $machine_key)
                 .map_err(|e| {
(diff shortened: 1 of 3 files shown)


dmulder (dismissed):
Need to review the copilot findings, but otherwise LGTM.
Next we need to go through each lock, and minimize the amount of time those locks are held.

dmulder:
@moritzheiber would you review my changes to your branch?

moritzheiber:
@dmulder LGTM! I think we ought to fix those two last remarks from Copilot?

dmulder:
I got a little carried away and started reworking paths to only hold mutex locks as little time as possible. There are other things that could be improved, IMHO, but I think this is a good start.

PR #1311, opened 2026-04-13
RPM pam-himmelblau, fix authselect and only amend PAM on install

FingerlessGlov3s:
Fixes #
- awk pattern for getting the authselect features was incorrect, so the select command would error trying to swap it to select himmelblau
- Whether authselect ran or not, the `aad-tool configure-pam` would always run, and make changes to the PAM config, so that now only runs if `authselect` or `pam-config` couldn't be used.
- I also then made it so it only does the PAM configuration on the initial rpm package install, this means when the system administrator makes changes to their PAM config after pam-himmelblau is installed, upgrading the package won't keep reverting them

Changes have been tested on a fresh install of Fedora 43

## Required Before Review

- [x] I manually tested this change on a test VM and confirmed it works as intended.
- [x] I listed exactly what testing I performed (commands/steps and observed results).
- [x] I listed which distro(s) and version(s) I tested on.
- [x] I ran relevant automated checks (for example `make test`, distro package build target, and `make test-selinux` when applicable) or explained why a check was not run.
- [x] If this change affects system integration (systemd, PAM/NSS/authselect, SELinux/AppArmor, filesystem paths, or credentials), I documented the packaging/runtime impact and any generator/template sources updated.
I don't believe I need to change any documentation here?
- [x] I linked the related issue/enhancement/discussion and confirmed the PR scope is focused and reviewable.

### DIFF
--- a/src/pam/Cargo.toml
+++ b/src/pam/Cargo.toml
@@ -54,11 +54,20 @@ if [ ! -e /lib64/security/pam_himmelblau.so ]; then
     ln -s /usr/lib64/security/pam_himmelblau.so /lib64/security/pam_himmelblau.so
 fi
 
+# Skip PAM configuration on upgrade ($1 >= 2) so admin customizations to
+# /etc/pam.d/*, authselect profiles, or pam-config state aren't reset by
+# a package update. Only the initial install wires PAM up.
+if [ "${1:-1}" -le 1 ]; then
+
 # 1) authselect first (if available)
+authselect_ok=0
 if command -v authselect >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


src/pam/Cargo.toml
```diff
@@ -91,11 +100,8 @@ fix_pam_config_account_line() {
 # 2) pam-config second (if available)
 pam_config_ok=0
 if command -v pam-config >/dev/null 2>&1; then
-    # Attempt to add himmelblau via pam-config. Older pam-config may not recognize --himmelblau at all.
```
bluca, line 94: why are these comments being removed?

FingerlessGlov3s: My bad, I've just forced pushed, only making necessary changes

I added lots of debugging prints, got carried way taking them out

bluca (approved):

PR #1260, opened 2026-03-17
fix: offline auth hang + decouple PAM socket timeout

bash-bandicoot:
## Summary
- **Commit 1:** Fixes offline auth hang by reducing `DEFAULT_CONN_TIMEOUT`
  from 30s to 5s for faster online detection. Also fixes slow online PIN
  login caused by unnecessary network probes.
- **Commit 2:** Decouples the PAM-to-daemon Unix socket timeout from the
  HTTP connection timeout. Introduces `DEFAULT_UNIX_SOCK_TIMEOUT = 90s`
  as a fixed constant, independent of `connection_timeout`. This prevents
  PAM_IGNORE errors when MFA approval + daemon background tasks (Kerberos,
  photo, Intune) exceed the socket timeout.

## Problem
When `connection_timeout` was lowered to 5s, the PAM socket timeout
(`connection_timeout * 2 = 10s`) became too short for MFA flows. The
daemon runs background tasks before sending the PAM response, easily
exceeding 10s and causing BrokenPipe / PAM_IGNORE on the client.

## Dependency
Requires the corresponding libhimmelblau fix to add timeouts to bare
reqwest clients in discovery code, otherwise discovery calls may hang
indefinitely with the reduced connection timeout:
https://gitlab.com/samba-team/libhimmelblau/-/merge_requests/147

## Files changed
- `src/common/src/constants.rs` — `DEFAULT_CONN_TIMEOUT = 5`,
  new `DEFAULT_UNIX_SOCK_TIMEOUT = 90`
- `src/common/src/config.rs` — `get_unix_sock_timeout()` returns fixed
  constant; updated tests

## Test plan
- [ ] Offline: verify auth fails fast (~5s) instead of hanging 30s
- [ ] Online with MFA: verify no PAM_IGNORE after successful MFA approval
- [ ] Online with Hello PIN: verify PIN login doesn't hang on network probe
- [ ] `sudo su` with Hello PIN: verify no socket timeout

PR #1364, opened 2026-05-05
fix(auth): apply connection_timeout to attempt_online() HTTP probes

bash-bandicoot:
## Summary

`attempt_online()` in both the himmelblau and OIDC providers used bare `reqwest::get()` with no timeout. When the network is down, that call blocks on Linux's default TCP timeout (minutes) instead of the configured `connection_timeout`, stalling auth flows that gate on the online check.

This PR builds a `reqwest::Client` with `connect_timeout` and total request `timeout` set to `connection_timeout`, and uses it for the probes.

| File | Probe |
|---|---|
| `src/common/src/idprovider/openidconnect.rs` | authorization_endpoint and openid-configuration fallback URL |
| `src/common/src/idprovider/himmelblau.rs` | authority_host probe |

Follow-up to closed PR #1260, scoped down to only the bug that is still novel after recent main changes (per-read PAM socket timeout in 41c3b41c, RwLock to Mutex refactor, decoupled Hello PRT deadlock fix in d97fa9cc).

## Test plan

| Step | Expected |
|---|---|
| Build packages for ubuntu24.04 (amd64 + arm64) | succeeds |
| Test VM with network reachable: PAM auth | succeeds, no regressions in journal |
| Test VM with network blackholed to authority host: PAM auth | fails fast within `connection_timeout` seconds, no multi-minute hang |

## Required Before Review

- [ ] I manually tested this change on a test VM and confirmed it works as intended.
- [ ] I listed exactly what testing I performed (commands/steps and observed results).
- [ ] I listed which distro(s) and version(s) I tested on.
- [ ] I ran relevant automated checks (for example `make test`, distro package build target, and `make test-selinux` when applicable) or explained why a check was not run.
- [ ] If this change affects system integration (systemd, PAM/NSS/authselect, SELinux/AppArmor, filesystem paths, or credentials), I documented the packaging/runtime impact and any generator/template sources updated.
- [ ] I linked the related issue/enhancement/discussion and confirmed the PR scope is focused and reviewable.

### DIFF
--- a/src/common/src/idprovider/himmelblau.rs
+++ b/src/common/src/idprovider/himmelblau.rs
@@ -28,6 +28,7 @@ use crate::constants::EDGE_BROWSER_CLIENT_ID;
 use crate::constants::ID_MAP_CACHE;
 use crate::db::KeyStoreTxn;
 use crate::idmap_cache::StaticIdCache;
+use crate::idprovider::common::build_online_probe_client;
 use crate::idprovider::common::flip_displayname_comma;
 use crate::idprovider::common::KeyType;
 use crate::idprovider::common::RefreshCacheEntry;
@@ -4252,7 +4253,21 @@ impl HimmelblauProvider {
             .authority_host()
             .await
             .unwrap_or(self.config.lock().await.get_authority_host(&self.domain));
-        match reqwest::get(format!("https://{}", authority_host)).await {
+        let request_timeout = self.config.lock().await.get_request_timeout();
+        let client = match build_online_probe_client(request_timeout) {
+            Ok(c) => c,
+            Err(e) => {
+                error!(?e, "Failed to build HTTP client for online check");
+                let mut state = self.state.lock().await;
+                *state = CacheState::OfflineNextCheck(now + OFFLINE_NEXT_CHECK);
+                return false;
+            }
+        };
+        match client
+            .get(format!("https://{}", authority_host))
+            .send()
+            .await
+        {
             Ok(resp) => {
                 if resp.status().is_success() {
                     debug!("provider is now online");
--- a/src/common/src/idprovider/openidconnect.rs
+++ b/src/common/src/idprovider/openidconnect.rs
@@ -20,6 +20,7 @@ use crate::config::HimmelblauConfig;
 use crate::constants::ID_MAP_CACHE;
 use crate::db::KeyStoreTxn;
 use crate::idmap_cache::StaticIdCache;
+use crate::idprovider::common::build_online_probe_client;
 use crate::idprovider::common::flip_displayname_comma;
 use crate::idprovider::common::KeyType;
 use crate::idprovider::common::TotpEnrollmentRecord;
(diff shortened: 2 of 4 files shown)


src/common/src/idprovider/himmelblau.rs
```diff
@@ -4222,7 +4175,21 @@ impl HimmelblauProvider {
             .authority_host()
             .await
             .unwrap_or(self.config.lock().await.get_authority_host(&self.domain));
-        match reqwest::get(format!("https://{}", authority_host)).await {
+        let conn_timeout = self.config.lock().await.get_connection_timeout();
```
dmulder: You should derive the connection timeout from the `request_timeout` parameter here instead (see how this is done in https://gitlab.com/samba-team/libhimmelblau/-/blob/main/src/auth.rs?ref_type=heads#L1531). `connection_timeout` makes more sense, but unfortunately has been used incorrectly for so long, that I've simply redefined it to refer to it's actual use case. We should perhaps rename `connection_timeout` to a more appropriate name, and simply have a un-documented mapping to the new parameter name.

dmulder: So essentially instead:

```rust
let request_timeout = self.config.lock().await.get_request_timeout();
let conn_timeout = min(request_timeout / 2, Duration::from_secs(3));
```

src/common/src/idprovider/openidconnect.rs
```diff
@@ -869,8 +861,23 @@ impl OidcProvider {
                 }
             };
 
+        let conn_timeout = self.config.lock().await.get_connection_timeout();
```
dmulder: Again:

```rust
let request_timeout = self.config.lock().await.get_request_timeout();
let conn_timeout = min(request_timeout / 2, Duration::from_secs(3));
```

dmulder (changes requested):

src/common/src/idprovider/himmelblau.rs
bash-bandicoot: Done in 16bdcefe. Switched both probes to use `request_timeout` and the libhimmelblau pattern: `connect_timeout = min(request_timeout / 2, 3s)`. Extracted the builder into `build_online_probe_client()` in `idprovider/common.rs` so the two providers share a single source of truth.

src/common/src/idprovider/openidconnect.rs
bash-bandicoot: Same change applied on this side. Now goes through `build_online_probe_client()` in `idprovider/common.rs`, which uses the `request_timeout / 2` capped at 3s pattern from libhimmelblau `auth.rs`.

src/common/src/idprovider/openidconnect.rs
```diff
@@ -861,8 +861,23 @@ impl OidcProvider {
...
+        let client = match reqwest::Client::builder()
+            .connect_timeout(Duration::from_secs(conn_timeout))
+            .timeout(Duration::from_secs(conn_timeout))
+            .build()
```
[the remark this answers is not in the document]
bash-bandicoot: Good call. Extracted into `build_online_probe_client()` in `src/common/src/idprovider/common.rs` in 16bdcefe, so both providers go through the same helper now. Future tweaks land in one place.

src/common/src/idprovider/himmelblau.rs
```diff
@@ -4175,7 +4175,21 @@ impl HimmelblauProvider {
...
+        let client = match reqwest::Client::builder()
+            .connect_timeout(Duration::from_secs(conn_timeout))
+            .timeout(Duration::from_secs(conn_timeout))
+            .build()
+        {
```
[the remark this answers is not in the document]
bash-bandicoot: Same fix as on the OIDC side. Both `attempt_online()` probes now go through `build_online_probe_client()` in `idprovider/common.rs`.

dmulder (dismissed):
LGTM

PR #1442, opened 2026-06-08
Fix systemd-user PAM override packaging

dmulder:
## Summary

Install a managed systemd-user PAM override when
the vendor openSUSE configuration would reject
NSS-backed users through pam_unix account checks,
and remove that override only during final
package uninstall.

## Testing

- **Distro(s) tested:**
- **Steps performed:**
- **Results observed:**

## Automated Checks

## Checklist


dmulder:
This is a backport from main (nightly).

dmulder:
We want to backport the nss shadow fix instead.

PR #1443, opened 2026-06-08
Run Intune policy checks asynchronously and on a user timer

### LINKED ISSUE #1435: Cache/throttle Intune policy application and run it async — `apply_policy` adds ~6 s to every PAM auth
## Summary

With `apply_policy = true`, the full Intune policy application runs synchronously on every
PAM authentication (GDM login, lock-screen unlock, every fresh sudo). On my machine that is
~6 s per auth, every time, even when the previous auth was seconds earlier. Suggesting two
independent improvements: throttle/cache the policy run, and detach it from the auth path.

## Problem statement

Environment: himmelblau 3.1.6, Ubuntu 24.04, Entra-joined + Intune enrolled,
`apply_policy = true`.

Measured on a fresh sudo (journal timestamps):

- `pam_himmelblau` init + PIN prompt: ~51 ms
- after PIN entry the tasks daemon runs KerberosConfig, KerberosTGTs, LoadProfilePhoto, ApplyPolicy
- ApplyPolicy alone: ~6 s; total daemon work after PIN entry: ~7 s
- network is not the bottleneck: 90–160 ms TTFB to the Microsoft endpoints from the same machine

So every login, unlock and fresh sudo ends up around 10–15 s wall clock including typing the
PIN. The auth only completes after ApplyPolicy finishes.

From reading `src/policies/src/policies.rs` (measured on 3.1.6; code re-checked against
current `main`):

- `apply_intune_policy()` re-runs the full sequence on every auth — the Intune round-trips,
  CSE execution, the status report, then the compliance fetch. I couldn't find any caching or
  throttling of the result, on 3.1.6 or on `main`.
- There is a hardcoded `sleep(Duration::from_secs(3))` before `get_compliance_info()` — a
  fixed 3 s on every auth. (On 3.1.6 it was a blocking `std::thread::sleep`; `main` already
  converts it to `tokio::time::sleep().await` in 763d2109 — thanks — but the fixed 3 s per
  auth remains.)

The tempting workaround is `apply_policy = false`, but that stops compliance reporting —
exactly what you don't want once Conditional Access requires a compliant device. (Browser SSO
presents the device claim but doesn't refresh compliance; the PAM-auth policy run is the only
thing that does.)

## Proposed solution

Two independent improvements, ideally both:

1. **Throttle/cache** — skip the full re-apply when policies were applied successfully within
   a configurable window (e.g. `apply_policy_interval`, default a few hours). MDM platforms
   typically check in a handful of times per day; once per PAM auth is far beyond what Intune
   itself expects.
2. **Async** — don't gate the PAM result on policy application. Queue the ApplyPolicy task
   after a successful auth and let it finish in the background; the auth decision doesn't
   depend on it.

Either one alone also retires the fixed 3 s pre-compliance-check delay from the auth path
(throttled runs skip it; async runs hide it). If it's load-bearing on the Intune side, a short
comment on why 3 s would help future readers.

## Benefits

- Login/unlock/sudo drop from ~7 s post-credential to sub-second, without losing compliance
  reporting.
- Removes the incentive to set `apply_policy = false` or to route around PAM, which is worse
  for fleet compliance than any throttling would be.
- Less per-auth load on the Intune endpoints.

## Alternatives considered

- `apply_policy = false` — fast, but stops compliance reporting.
- sudo `Defaults timestamp_timeout` bump — helps sudo only, not login/unlock.
- A local-password-first PAM stack — works, but compliance then depends on the user
  remembering to do regular PIN auths.

## Additional context

Related but distinct: #895 (offline mode after IPv6-only probe failure), #1392 (NSS
GetCredentialType throttle). Happy to test patches — this machine reproduces the timings
consistently.

---

### 💡 Help Make It Happen!

For US tax exempt donations:

dmulder:
## Summary

  This updates Intune policy application so PAM login no longer waits on the policy check path. Login-triggered policy application now runs asynchronously, is tracked per account, and is throttled to avoid repeated runs within a 30-minute window or concurrent checks for the same user.

  It also adds packaged systemd user units for periodic compliance checks:

  - Adds himmelblau-compliance-check.service
  - Adds himmelblau-compliance-check.timer
  - Runs aad-tool compliance-check on boot after 10 minutes and then every 8 hours
  - Enables the timer globally during install and disables it on package removal
  - Installs the user units for Debian and openSUSE packages
  - Adds unit tests for the Intune policy throttle behavior

  Testing: not run; this was a diff review only.

Fixes #1435

## Testing

- **Distro(s) tested:**
- **Steps performed:**
- **Results observed:**

## Automated Checks

## Checklist


src/daemon/scripts/postinst
```diff
@@ -116,6 +116,11 @@ if command -v deb-systemd-helper >/dev/null 2>&1; then
             deb-systemd-invoke restart himmelblaud.service himmelblaud-tasks.service >/dev/null || true
         fi
     fi
+
+    # Globally enable user timer for compliance checks (does not start existing sessions)
+    if [ -z "$DPKG_ROOT" ] && [ -d /run/systemd/system ]; then
+        systemctl --global enable himmelblau-compliance-check.timer 2>/dev/null || true
```
bluca: please use deb-systemd-helper (with --user) as above

src/daemon/scripts/postrm
```diff
@@ -26,6 +26,11 @@ if [ "$_is_full_remove" = "1" ]; then
         systemctl disable himmelblaud.service himmelblaud-tasks.service himmelblau-hsm-pin-init.service >/dev/null || true
     fi
 
+    # Globally disable and remove user timer on final package removal
+    if command -v systemctl >/dev/null 2>&1; then
+        systemctl --global disable himmelblau-compliance-check.timer 2>/dev/null || true
```
bluca: as above

PR #1457, opened 2026-06-17
Make Kerberos cache and Profile picture optional

Srylax:
## Summary

Currently, the `PamAuthenticateStep` fetches Kerberos TGT every time even if it fails. The same thing happens with the profile picture. This adds additional burden for the PAM login and adds noise to the log. 

## What Changed

Adds two no option `enable_kerberos_cache` and `fetch_profile_picture` these can be 

## Testing

- **Distro(s) tested:** NixOS
- **Steps performed:** PAM authentication
- **Results observed:** login is faster

## Automated Checks

- [x] I ran `make test` (or explained why not)
- [x] I ran the distro package build target (or explained why not) -> yes for NixOS
- [~] I ran `make test-selinux` (if applicable)

## Checklist

- [X] I manually tested this change on a test VM
- [~] PR scope is focused and linked to an issue/enhancement/discussion


shell.nix
```diff
@@ -18,6 +18,9 @@ pkgs.mkShell {
     stdenv.cc
     pam
     sqlite.dev
+    python3
+    rustc
+    cargo
```
dmulder, line 23: What is the purpose of these? Your commit just says that you're adding them, but doesn't explain why.

dmulder:
Overall, I think this is a very good idea. In addition to this, I think these should be disabled by default when we configure OIDC auth (since they're creating a bunch of log noise for OIDC). I can handle that with a follow up PR.

PR #1460, opened 2026-06-17
Make MFA poll prompt services configurable

### LINKED ISSUE #1453: Two digit Authenticator App code not displayed in Cockpit
### Description

I am using Debian 13 with Cockpit installed.

When logging in to Cockpit I get the push notification in the Microsoft Authenticator App but the two numbers needed are not displayed in Cockpit.

I believe this is the same issue as OpenSSH Bug 2876 which is resolved by adding mfa_poll_prompt to the PAM config.

However looking at the code it appears mfa_poll_prompt only triggers if the service name contains ssh.

src/common/src/auth.rs

_// Only prompt if this is the ssh service and a message was sent.
if state.opts.mfa_poll_prompt && state.service.contains("ssh") && !msg.trim().is_empty() {
    state.msg_printer.prompt_echo_off("Press enter to continue");
}_

Would it be possible to add cockpit to this list or make what services should prompt configurable in the himmelblau.conf file?

### Steps to Reproduce

1. Open Cockpit in a browser
2. Enter valid Username and Password
3. Session will time out when using Authenticator Push as code is not displayed to screen.

Expectation

The two digit code to be displayed so it can be entered into the App 

### Relevant log output

```shell

```

### Version

himmelblaud 3.1.6

### Code of Conduct

---

### 💡 Help Make It Happen!

For US tax exempt donations:

dmulder:
## Summary

Add a generated mfa_poll_prompt_services option so Cockpit can use the same PAM prompt-flush workaround as SSH. The auth path now checks the configured service list, filters empty string-list values, and documents the new option.

Fixes #1453

## Testing

- **Distro(s) tested:**
- **Steps performed:**
- **Results observed:**

## Automated Checks

## Checklist

### DIFF
--- a/man/man5/himmelblau.conf.5
+++ b/man/man5/himmelblau.conf.5
@@ -1,4 +1,4 @@
-.TH HIMMELBLAU.CONF "5" "May 2026" "Himmelblau Configuration" "File Formats"
+.TH HIMMELBLAU.CONF "5" "June 2026" "Himmelblau Configuration" "File Formats"
 .SH NAME
 himmelblau.conf \- Configuration file for Himmelblau, enabling Azure Entra ID authentication on Linux.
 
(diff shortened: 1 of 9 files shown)


dmulder:
@paul-turton-taxcalc Can you test this fix?

paul-turton-taxcalc:
> @paul-turton-taxcalc Can you test this fix?

@dmulder Thanks for looking at this so quicky. I have tested on Debian 13 and can confirm it has solved the problem. I have also tried overriding the default mfa_poll_prompt_services in the config file and removing/adding services produces the results you would expect :)

PR #1502, opened 2026-07-04
chore(deps): update libwebauthn to 0.9.0

AlfioEmanueleFresta:
Thanks for using [libwebauthn](https://github.com/linux-credentials/libwebauthn)! I maintain it, so I figured I'd save you the bump. 0.3.0 is six releases behind. It's picked up a lot of security fixes since then, the API surface has changed non-trivially, and a pile of bugs got cleared out along the way, including caBLE fixes around Android compatibility.

The only file that touches libwebauthn is `src/common/src/auth.rs` (the caBLE QR login), so the migration is small. Three breaking-API adjustments:

- `GetAssertionRequest`: `cross_origin` is replaced by `top_origin`. himmelblau sets neither, so it stays `None`.
- `CableQrCodeDevice::new_transient` now takes a `CableTransports`. I pass `CloudAssistedOnly`, which preserves 0.3.0's exact behavior (the alternative also advertises a direct BLE L2CAP channel, which isn't stable yet).
- `Device::channel` now takes a `ChannelSettings`. I pass `ChannelSettings::default()`, which is empty today. Its only field is an optional store for reusing PIN/UV tokens across sessions, which himmelblau doesn't use.

I compiled the migration against 0.9.0 locally, but will let CI cover the full build. The one thing CI can't exercise is the live phone leg, so could one of you smoke-test a caBLE phone login before merging? Happy to jump in if anything looks off. Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -642,6 +642,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "bluer"
+version = "0.17.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "af68112f5c60196495c8b0eea68349817855f565df5b04b2477916d09fb1a901"
+dependencies = [
+ "futures",
+ "hex",
+ "libc",
+ "log",
+ "macaddr",
+ "nix",
+ "num-derive",
+ "num-traits",
+ "serde",
+ "serde_json",
+ "strum",
+ "tokio",
+ "uuid",
+]
+
+[[package]]
 name = "bluez-async"
 version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


dmulder:
The reason I haven't updated yet is because the latest version introduces some dependency changes which I haven't had the chance to vet properly (specifically via `bluer`, which has some large un-vetted dependencies). We can keep this PR open, but those new dependencies need community review.

dmulder:
> I compiled the migration against 0.9.0 locally, but will let CI cover the full build. The one thing CI can't exercise is the live phone leg, so could one of you smoke-test a caBLE phone login before merging? Happy to jump in if anything looks off. Thanks!

I'll post this to the community and see if someone can test this out. @ccadruvi you added this integration originally, would you test this out before we merge it?

ccadruvi:
A colleague tested it with this PR, looks good to us!

dmulder (dismissed):
LGTM. Getting ready to merge.

dmulder (approved):

PR #1536, opened 2026-07-20
fix(broker): try silent token acquire before interactive pinentry

bash-bandicoot:
## Summary

- Edge/OneAuth on Linux calls D-Bus `AcquireTokenInteractively` (`com.microsoft.identity.broker1`) even when a warm PRT can already satisfy the request.
- `InteractiveSessionBroker::acquire_token_interactively` previously **always** ran pinentry interactive auth first, then silent. Interactive auth requires `DISPLAY`/`WAYLAND_DISPLAY` + pinentry in the **broker process**; `himmelblau-broker` often runs under `background.slice` without those vars.
- Result: immediate `org.freedesktop.DBus.Error.Failed` → Edge “Something went wrong” / tag **4ulu3**, despite `getAccounts` working and a usable PRT.

This PR prefers **silent-first**, matching the existing silent-then-interactive pattern used for PRT SSO cookies:

1. Try `acquire_token_silently`
2. Only if that fails **and** `interactive_session_available()`: pinentry re-auth, then silent again
3. Otherwise return a clear D-Bus error (instead of failing interactive when silent would have worked)

## Why this is Edge-visible

Chrome on Himmelblau typically uses `himmelblau-sso` / native messaging for web SSO. Edge profile “Sign in to sync” uses OneAuth → broker Interactive API, so it hits this path.

## Test plan

- [ ] Build/install patched `himmelblau-broker` on an Entra-joined host with warm PRT
- [ ] Confirm broker env often lacks `DISPLAY`/`WAYLAND` (`tr '\\0' '\\n' < /proc/$(systemctl --user show -p MainPID --value himmelblau-broker.service)/environ | grep -E DISPLAY|WAYLAND`)
- [ ] Edge → Sign in to sync / work account: should succeed (or at least not fail immediately with 4ulu3 when silent token works)
- [ ] Cold PRT / expired session: silent fails → interactive prompt only when a graphical session is available
- [ ] Regression: `getAccounts`, `acquireTokenSilently`, PRT SSO cookie path still behave as before

### DIFF
--- a/src/broker/src/main.rs
+++ b/src/broker/src/main.rs
@@ -169,9 +169,10 @@ impl MessagePrinter for PinentryMessagePrinter {
 }
 
 /// Session broker that forwards all D-Bus calls to the himmelblaud
-/// broker socket, but overrides `acquireTokenInteractively` to
-/// drive a full interactive auth (via `authenticate()` on the main
-/// daemon socket with pinentry) when silent acquisition fails.
+/// broker socket, but overrides `acquireTokenInteractively` to try
+/// silent acquisition first and only fall back to interactive auth
+/// (via `authenticate()` on the main daemon socket with pinentry)
+/// when silent fails and a graphical session is available.
 struct InteractiveSessionBroker {
     /// Path to the himmelblaud broker socket.
     broker_sock_path: String,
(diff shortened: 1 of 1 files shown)


bash-bandicoot:
@dmulder , please check when you have time

dmulder:
@bluca do you think this would interfere with your current workflow? Otherwise I think I'm in favor of this change.

bluca:
I don't think this makes sense? Interactive means interactive. It will cause failures as the provided token won't have the necessary requisites (eg: some apps require a fresh token or fail otherwise, like the vpn auth), even though it's an otherwise valid and not-expired token.

dmulder:
> I don't think this makes sense? Interactive means interactive. It will cause failures as the provided token won't have the necessary requisites (eg: some apps require a fresh token or fail otherwise, like the vpn auth), even though it's an otherwise valid and not-expired token.

Maybe the correct question is, why didn't the user's system detect a graphical environment for the re-auth when trying to request a token for Edge? Instead of a band-aid such as the silent auth here, perhaps we aught to determine why it failed and fix that instead.

bash-bandicoot:
`Maybe the correct question is, why didn't the user's system detect a graphical environment for the re-auth when trying to request a token for Edge? Instead of a band-aid such as the silent auth here, perhaps we aught to determine why it failed and fix that instead.`


On the failing host the user did have a graphical session; Edge still hit Interactive and failed. My theory is that himmelblau-broker (user service / background.slice) often lacks DISPLAY/WAYLAND_DISPLAY in its own environment, so interactive_session_available() returns false and pinentry never runs, even though the desktop session has a display.

dmulder:
> On the failing host the user did have a graphical session; Edge still hit Interactive and failed.

I'm confused. How could the user have the Edge browser, but no graphical session?

bash-bandicoot:
The user did have a graphical session (Edge + Wayland/Xwayland). I’m not claiming “no GUI.”

The open question is why himmelblau-broker didn’t treat the session as interactive: it only checks DISPLAY/WAYLAND_DISPLAY in its own process environment.

I’ll gather that evidence and follow up.

bash-bandicoot:
I was misled: quick Interactive failures + broker under background.slice made me assume missing DISPLAY/WAYLAND, so I proposed silent-first. That was wrong.

Agree with @bluca Interactive shouldn’t be weakened for that. Closing #1536.

PR #1540, opened 2026-07-22
pam: fold info text into the password prompt (single PAM_PROMPT_ECHO_…

### LINKED ISSUE #1539: PAM_TEXT_INFO before the password prompt breaks MariaDB auth_pam
### Description

During PAM authentication, pam_himmelblau emits a standalone PAM_TEXT_INFO message (the configured entra_id_password_prompt, and long_prompt when present) immediately before the PAM_PROMPT_ECHO_OFF password prompt.

Some PAM clients cannot handle an info message at that point in the conversation. MariaDB's auth_pam plugin relays the PAM conversation to the client over its dialog protocol, and the extra PAM_TEXT_INFO makes the relay abort with:
```
auth_pam_tool: pam_unix(mariadb:auth): unexpected response from failed conversation function
```
The password prompt never reaches the client, so Entra-backed PAM authentication to MariaDB is impossible. A plain pam_unix stack in the same PAM service works (it sends a single PAM_PROMPT_ECHO_OFF and nothing else), which isolates the incompatibility to the info-message-before-prompt pattern rather than anything MariaDB-specific in the credentials or SSL path. OpenSSH is affected more mildly — info messages during authentication are not reliably displayed (OpenSSH bug 2876) — so the text isn't dependable there either.

### Steps to Reproduce

1. On a himmelblau-joined host, create a PAM service /etc/pam.d/mariadb whose auth stack invokes pam_himmelblau.
2. Enable the MariaDB auth_pam plugin and create a user identified through it, e.g.:
```
INSTALL SONAME 'auth_pam';
CREATE USER 'someuser'@'%' IDENTIFIED VIA pam USING 'mariadb';
```
3. From a MariaDB client, connect as that user so the server invokes the mariadb PAM service:
```
mysql -u someuser -p
```
4. Observe: the client never receives the password prompt; the connection fails, and the server log shows auth_pam_tool: pam_unix(mariadb:auth): unexpected response from failed conversation function.

Expected: the password prompt reaches the client and PAM authentication proceeds (as it does with a pam_unix-only stack).
Actual: the conversation aborts before the prompt, because of the leading PAM_TEXT_INFO.

### Relevant log output

```shell

```

### Version

3.1.9

### Code of Conduct

---

### 💡 Help Make It Happen!

For US tax exempt donations:

ximpanz:
## Summary

pam_himmelblau sends a standalone PAM_TEXT_INFO before the password prompt.
Some PAM clients don't honour info messages during authentication (OpenSSH
bug 2876), and for others it breaks the conversation outright: MariaDB's
auth_pam dialog relay aborts with "unexpected response from failed
conversation function", making Entra-backed PAM authentication to MariaDB
impossible. This folds the info text into the prompt message itself, so the
conversation carries a single PAM_PROMPT_ECHO_OFF and the user still sees all
the text.

Fixes #1539

## What changed

- `handle_pam_auth_response_password` no longer emits standalone PAM_TEXT_INFO:
  `long_prompt` and the configured `entra_id_password_prompt` are folded into
  the single PAM_PROMPT_ECHO_OFF prompt message.
- Test support: `RecordingPrinter` now records prompts and returns a
  credential; two new unit tests assert the behavior (no PAM_TEXT_INFO emitted,
  one prompt containing all the text, credential flows through).

## Testing

- **Distro(s) tested:** Debian 12, Debian 13
- **Steps performed:** `make test`; built the Debian packages (`make debian12`,
  `make debian13`) and installed them on a test host; connected to MariaDB
  (10.11, `auth_pam` + pam_himmelblau in the `mariadb` PAM service).
- **Results observed:** the password prompt now reaches the client and the PAM
  conversation proceeds where it previously aborted with "unexpected response
  from failed conversation function"; SSH/sudo prompts behave as before, with
  the info text delivered inside the prompt message.

## Automated Checks

- [x] I ran `make test` (all crates green, incl. 2 new tests for this change)
- [x] I ran the distro package build target (`make debian12`, `make debian13`)
- [x] I ran `make test-selinux` — no SELinux changes; policy passes on
  rocky8/9/10, fedora43/44, tumbleweed; sle16 skipped (needs SUSE SCC creds)

## System Integration Impact

PAM only. The password branch's conversation changes from
PAM_TEXT_INFO + PAM_PROMPT_ECHO_OFF to a single PAM_PROMPT_ECHO_OFF whose
message includes the info text. No NSS, systemd, SELinux/AppArmor, filesystem
path, or credential handling changes.

## Checklist

- [x] I manually tested this change on a test VM
- [x] PR scope is focused and linked to an issue/enhancement/discussion

### DIFF
--- a/src/common/src/auth.rs
+++ b/src/common/src/auth.rs
@@ -822,21 +822,36 @@ fn handle_pam_auth_response_password(
     let cred = if let Some(cred) = consume_authtok {
         cred
     } else {
-        if let Some(long_prompt) = long_prompt.filter(|prompt| !prompt.trim().is_empty()) {
-            state.msg_printer.print_text(long_prompt);
+        let fold = state.opts.no_info_prompt;
+        let long_prompt = long_prompt.filter(|prompt| !prompt.trim().is_empty());
+
+        // The info text is normally sent as a standalone PAM_TEXT_INFO before the
+        // prompt. With no_info_prompt it is folded into the single password prompt
+        // instead, for PAM clients that cannot carry a separate info message during
+        // authentication (e.g. MariaDB's auth_pam dialog).
+        if !fold {
+            if let Some(long_prompt) = long_prompt {
+                state.msg_printer.print_text(long_prompt);
+            }
         }
 
         let prompt = match prompt.filter(|prompt| !prompt.trim().is_empty()) {
             Some(prompt) => i18n::translate_external_message(prompt),
             None if state.cfg.get_oidc_issuer_url().is_some() => tr("Cloud Password:"),
             None => {
-                state
-                    .msg_printer
-                    .print_text(&i18n::translate_external_message(
-                        &state.cfg.get_entra_id_password_prompt(),
-                    ));
-                tr("Entra Id Password:")
+                let info =
+                    i18n::translate_external_message(&state.cfg.get_entra_id_password_prompt());
+                if fold {
+                    format!("{}\n{}", info, tr("Entra Id Password:"))
+                } else {
+                    state.msg_printer.print_text(&info);
+                    tr("Entra Id Password:")
+                }
             }
+        };
+        let prompt = match (fold, long_prompt) {
+            (true, Some(long_prompt)) => format!("{}\n{}", long_prompt, prompt),
+            _ => prompt,
         };
         match state.msg_printer.prompt_echo_off(&prompt) {
             Some(cred) => cred,
@@ -1563,6 +1578,7 @@ mod tests {
     struct RecordingPrinter {
         text: Mutex<Vec<String>>,
         error: Mutex<Vec<String>>,
+        prompts: Mutex<Vec<String>>,
     }
 
     impl MessagePrinter for RecordingPrinter {
@@ -1578,8 +1594,9 @@ mod tests {
             None
         }
 
-        fn prompt_echo_off(&self, _prompt: &str) -> Option<String> {
-            None
+        fn prompt_echo_off(&self, prompt: &str) -> Option<String> {
+            self.prompts.lock().unwrap().push(prompt.to_string());
+            Some("hunter2".to_string())
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/common/src/auth.rs
```diff
@@ -1564,8 +1562,9 @@ mod tests {
             None
         }
 
-        fn prompt_echo_off(&self, _prompt: &str) -> Option<String> {
-            None
+        fn prompt_echo_off(&self, prompt: &str) -> Option<String> {
+            self.prompts.lock().unwrap().push(prompt.to_string());
+            Some("hunter2".to_string())
```
dmulder, line 1599: What is `hunter2`?

src/common/src/auth.rs
```diff
@@ -820,21 +820,18 @@ fn handle_pam_auth_response_password(
...
+            None => format!(
+                "{}\n{}",
+                i18n::translate_external_message(&state.cfg.get_entra_id_password_prompt()),
                 tr("Entra Id Password:")
-            }
+            ),
+        };
+        let prompt = match long_prompt.filter(|prompt| !prompt.trim().is_empty()) {
+            Some(long_prompt) => format!("{}\n{}", long_prompt, prompt),
+            None => prompt,
```
dmulder: This wraps the long prompt inside the short prompt, causing a hideous mess on GDM and other UXs.
Also, wouldn't this only fix the one obscure case you've described, and only when authenticating with just the password? What happens when you authenticate with MFA when mariadb auth is configured?

dmulder (commented):
I'm not sure that we can accept this patch. This change fixes one obscure use case, while breaking others.
I'd also push back, because this is technically a bug in MariaDB, not in Himmelblau. This patch is just a band-aid to fix poor PAM handling in MariaDB.

dmulder:
I think we need to engage with the MariaDB community on this one. Their code is assuming you only ever authenticate via pam_unix, and doesn't properly handle pam prompts.

dmulder:
@ximpanz I have an alternate proposal. What if we added an option which disabled the extra prompts (instead of folding them in)? This would fix this particular case (and potentially other issues), while not modifying the default behavior.

ximpanz:
Hi,

Ok, like the no_hello_pin option? If so I can rework this using an option.

Cheers Kalle

dmulder:
> Ok, like the no_hello_pin option? If so I can rework this using an option.

Yeah, kind of. Just add an option to `src/common/docs-xml/himmelblauconf/base/` and the related bits for using the option to turn off that extra prompt.

src/common/src/auth.rs
ximpanz: Just a string to use in the test, an old bash.org IRC joke :)

dmulder (dismissed):
LGTM

dmulder (approved):

PR #1547, opened 2026-07-28
Limit repeated try_unseal TPM unlock attempts

### LINKED ISSUE #1538: try_unseal adds ~23s to every interactive login (full online re-auth, ~40 sequential TPM ops)
### Description

On interactive login (GDM, and `sudo`) the `auth optional pam_himmelblau.so try_unseal` line in `common-auth` adds ~23s of latency after I type my password. It's `optional` so it doesn't change the login outcome, but PAM still waits for it, so every login (and every `sudo`) hangs for ~23s.

A debug trace (`debug = true`) shows the time is entirely inside `unix_user_online_auth_step`, and it's actually ~100% TPM crypto with near-zero network (the connections get reused). What surprised me is that a single unseal does a *full* online re-auth (looks like it exchanges the PRT for a new PRT and re-applies policy) and reloads the TPM keys over and over. Counting the spans for one login:

| op | count | total |
|---|---|---|
| `rs256_load` | 11 | 9.0s |
| `unseal_data` | 7 | 6.0s |
| `ms_hello_key_load` | 3 | 6.0s |
| `storage_key_load_pin` | 3 | 3.5s |
| `msoapxbc_rsa_decipher_session_key` | 3 | 2.6s |
| `seal_data` | 6 | 2.4s |
| `ms_device_enrolment_key_load` | 2 | 1.6s |

Trace excerpt:

```
pam try unseal [ 23.2s | 0.03% / 100.00% ]
  unix_user_online_auth_step [ 23.1s | 0.06% / 99.79% ]
    ms_hello_key_load [ 1.99s ]
      storage_key_load_pin [ 1.18s ]
      rs256_load [ 812ms ]
    ms_hello_key_load [ 1.99s ]
      storage_key_load_pin [ 1.18s ]
      rs256_load [ 814ms ]
    unseal_data [ 858ms ]
    rs256_load [ 816ms ]
    seal_data [ 395ms ]
    ... (many more rs256_load / unseal_data / seal_data)
```

Each individual TPM op is ~0.5-0.8s, which (comparing with #1164, where `rs256_load`/`unseal_data` were ~0.55s each on different hardware) doesn't look abnormal for a discrete TPM. So the cost here seems to come mostly from the *number* of ops (~40 sequential per login), not from my TPM being unusually slow. It's a discrete ST `ST33KTPM2XSPI` on a ThinkPad P1 Gen 7, in case that matters.

A few things I'm wondering (hedged, might be off base):

1. Could the loaded Hello / storage key be **cached in-memory for the duration of one auth**? The trace reloads `ms_hello_key_load` 3x and `rs256_load` 11x within a single `unix_user_online_auth_step`, and each reload pays the full `storage_key_load_pin` + `rs256_load` cost. Loading once and reusing looks like it could roughly halve the time (or better).
2. Does `try_unseal` need a **full online PRT refresh on every login**, or could it reuse still-valid cached material (within `cache_timeout`) and skip the exchange when the PRT hasn't expired? My logins are minutes-to-hours apart, so I pay the full cost each time.
3. Since it's `optional` and mostly there to keep the session/compliance fresh, would it make sense to run the unseal **non-blocking / async** so it doesn't add latency to the interactive login path?

### Environment

- himmelblau `4.0.0-ubuntu24.04~20260714`, Ubuntu 24.04
- `hsm_type = tpm`, discrete ST `ST33KTPM2XSPI` (fw 9.256.0.0), ThinkPad P1 Gen 7
- Two tenants configured, Hello PIN login
- Standard `Azure Entra Id unlock` PAM profile (`auth optional pam_himmelblau.so try_unseal` in `common-auth`)

---

### 💡 Help Make It Happen!

For US tax exempt donations:

dmulder:
## Summary

Fixes #1538

## What Changed

Adds an in-memory `try_unseal` result cache in the resolver to avoid repeatedly entering the expensive TPM-backed unlock path for the same account.

- Successful `try_unseal` results are cached for the existing resolver cache timeout.
- Failed or non-actionable `try_unseal` attempts are cached for a short 60 second cooldown.
- Auth init/step errors are also cooldown-gated so repeated PAM invocations do not immediately retry the same TPM-heavy flow.
- Added focused unit tests for success reuse, failure cooldown, and retry suppression after an error.

## Testing

- **Distro(s) tested:** Not run locally; workspace lacks a Rust toolchain and Docker daemon access.
- **Steps performed:** `git diff --check`; attempted `make test`.
- **Results observed:** `git diff --check` passed. `make test` failed because Docker could not connect to `/var/run/docker.sock`.

## Automated Checks

`make test` was attempted but could not run because the Docker daemon/socket is unavailable in this environment. No distro package build was run for the same reason. `make test-selinux` was not run; this change does not affect SELinux policy.

## System Integration Impact

Affects PAM `try_unseal` behavior. Repeated optional `try_unseal` calls for the same account are now bounded by an in-memory resolver cache: successful unseals are reused for the resolver cache timeout, and failures are suppressed for 60 seconds. This reduces repeated TPM unlock/key-load attempts during login/sudo loops without persisting additional credentials.

## Checklist

- [x] PR scope is focused and linked to an issue/enhancement/discussion

### DIFF
--- a/src/common/src/resolver.rs
+++ b/src/common/src/resolver.rs
@@ -52,6 +52,14 @@ use tokio::sync::broadcast;
 use himmelblau::auth::UserToken as UnixUserToken;
 
 const NXCACHE_SIZE: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(128) };
+const TRY_UNSEAL_CACHE_SIZE: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(128) };
+const TRY_UNSEAL_FAILURE_COOLDOWN_SECONDS: u64 = 60;
+
+#[derive(Debug)]
+enum TryUnsealCacheState {
+    SuccessUntil(SystemTime),
+    FailureUntil(SystemTime),
+}
 
 #[allow(clippy::large_enum_variant)]
 pub enum AuthSession {
@@ -95,6 +103,7 @@ where
     allow_id_overrides: HashSet<Id>,
     nxset: Mutex<HashSet<Id>>,
     nxcache: Mutex<LruCache<Id, SystemTime>>,
+    try_unseal_cache: Mutex<LruCache<String, TryUnsealCacheState>>,
 }
 
 impl Display for Id {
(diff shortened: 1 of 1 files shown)


src/common/src/resolver.rs
```diff
@@ -1764,34 +1840,87 @@ where
...
+    }
+
     /// One-shot PIN-based unseal: runs auth init and, if the daemon
     /// requests a PIN, immediately submits it. Returns Ok(true) on
     /// success, Ok(false) on auth failure, Err(()) on internal error.
     pub async fn pam_try_unseal(&self, account_id: &str, cred: &str) -> Result<bool, ()> {
+        let now = SystemTime::now();
+        if let Some(cached_result) = self.check_try_unseal_cache(account_id, now).await {
+            return Ok(cached_result);
+        }
```
dmulder, line 1887: This is an authentication bypass. Simply returning the cached state of the operation _does_ limit our TPM hits, but it also allows _anyone_ in that cache window to authenticate.

dmulder:
This is simply a bad attempt to fix the bug. Rejecting.

PR #1549, opened 2026-07-28
Reconcile local groups periodically

### LINKED ISSUE #719: Add/remove users to local groups between authentication/login attempts
**Summary**
Enable periodic check of group membership to reflect Azure group changes between login attempts.

**Problem Statement**
At the moment, himmelblau service will add users to the local groups based on the main configuration file. In case of support for "sudo" group, service will also remove user from the local sudo group. However, this happens only on service restart or when user logs out and in again. If administrator removes user from the Entra sudo group this change will reflect on the system only when user authenticates again, which might be "to late".

**Proposed Solution**
Implement periodic checks of Azure Entra ID group membership and keep the local sudoers file in sync.

**Benefits**
Entra group changes will reflect on the local system faster, depending on the Azure group sync frequency.

**Alternatives Considered**
This functionality can be supported by https://www.man7.org/linux/man-pages/man8/systemd-userdbd.service.8.html. For this to work, systemd nss module must be enabled in /etc/nsswitch.conf.

himmelbot:
Fixes #719

## Summary
- add enabled-by-default periodic reconciliation for configured local group membership
- reuse the existing root task-daemon LocalGroups path so sudo_groups revocation can be reflected between login attempts
- add local_groups_reconcile_interval config, defaulting to 300 seconds; set to 0 to disable
- document the new option in the man page and NixOS options

## systemd-userdb assessment
systemd-userdb is not a direct replacement for the current local group mutation path here. It can publish NSS-style user/group records through systemd-userdbd, but Himmelblau still owns the Entra membership decision and existing NSS integration. For the requested sudo_groups revocation behavior, periodically reconciling cached Entra users through the existing task daemon keeps behavior consistent with login-time reconciliation and avoids requiring operators to replace NSS/group handling with a separate userdb provider.

## Validation
- cargo fmt --all --check
- git diff --check
- generated Rust config accessor checked from XML

Full cargo check/test could not complete in this environment because libdbus-sys requires pkg-config/dbus-1 metadata and pkg-config is not installed.

### DIFF
--- a/src/daemon/src/daemon.rs
+++ b/src/daemon/src/daemon.rs
@@ -19,7 +19,7 @@
 #![deny(clippy::needless_pass_by_value)]
 #![deny(clippy::trivially_copy_pass_by_ref)]
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::error::Error;
 use std::fs::metadata;
 use std::io;
@@ -93,6 +93,8 @@ type AsyncTaskRequest = (TaskRequest, oneshot::Sender<TaskOutcome>);
 type IntunePolicyThrottle = Arc<Mutex<HashMap<String, IntunePolicyThrottleState>>>;
 
 const INTUNE_POLICY_THROTTLE_INTERVAL: Duration = Duration::from_secs(30 * 60);
+const LOCAL_GROUPS_TASK_QUEUE_TIMEOUT: Duration = Duration::from_millis(100);
+const LOCAL_GROUPS_TASK_RESPONSE_TIMEOUT: Duration = Duration::from_millis(1000);
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum IntunePolicyThrottleState {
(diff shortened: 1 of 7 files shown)


src/daemon/src/daemon.rs
```diff
@@ -244,6 +246,111 @@ async fn handle_task_client(
...
+    cachelayer: &Resolver<HimmelblauMultiProvider>,
+    task_channel_tx: &Sender<AsyncTaskRequest>,
+    cfg: &HimmelblauConfig,
+) {
+    if cfg.get_local_groups().is_empty() && cfg.get_sudo_groups().is_empty() {
+        trace!("Skipping local group reconciliation; no local groups configured");
+        return;
+    }
+
+    let accounts = match cachelayer.get_nssaccounts().await {
```
dmulder, line 333: `get_nssaccounts` doesn't refresh to the Entra ID groups of the user. It only pulls what we already know from the cache. This change therefore has no effect on the reported issue #719 because the original issue was that we only updated this information after a user logs in.
A true fix would require sending a query to the MS graph to re-check group information for every user, on the interval.

dmulder (changes requested):

dmulder (approved):
LGTM

dmulder (dismissed):

PR #1552, opened 2026-07-28
Handle generic OIDC username collisions by subject

### LINKED ISSUE #1335: Differentiate between similar usernames for generic OIDC based on sub e.g.
### Summary

OIDC claim preferred_username may not be unique, so use sub as identy pin.

### Problem Statement

In our ecosystem it is possible to receive the same `preferred_username` for different identities. The reliance on `preferred_username` is addressed in https://github.com/himmelblau-idm/himmelblau/issues/1326 but even then the operator may choose a claim that is not globally unique for posix username matching.

### Proposed Solution

The `sub` of both identities however is always globally unique but not convenient as a login name, so it would be nice if Himmelblau was able to match the received oidc username claim against a stored value for a previously connected `sub` (or any other configurable, stable, globally unique claim in the userinfo response).

### Benefits

Resolved Linux username would always be attached to a single OIDC identity.

### Alternatives Considered

_No response_

### Additional Context

Receiving the same login name for a different sub could create a new user, possibly extended with a _1 postfix?

### Code of Conduct

---

### 💡 Help Make It Happen!

Want to see this issue resolved faster? Fund its implementation through our **Backer's Bounty** program, where you choose which issues get priority!

For US tax exempt donations:

himmelbot:
## Summary
- use the OIDC subject as the stable generic OIDC identity key stored in spn
- derive generic OIDC UUID/ID mapping from that stable key
- avoid overwriting an existing local username owned by a different subject by assigning a deterministic suffixed local name
- add unit coverage for same-subject reuse and conflicting-subject suffixing

Fixes #1335

## Verification
- cargo fmt --check
- cargo test -p himmelblau_unix_common oidc_local_name --lib (blocked: build environment lacks pkg-config/dbus metadata required by libdbus-sys)

### DIFF
--- a/src/common/src/idprovider/openidconnect.rs
+++ b/src/common/src/idprovider/openidconnect.rs
@@ -1237,6 +1237,51 @@ mod tests {
 
         assert!(mapped.is_empty());
     }
+
+    #[test]
+    fn oidc_local_name_reuses_existing_name_for_same_identity() {
+        let object_id = Uuid::parse_str("0180c9c1-6123-7a21-b520-f839dc963d54").unwrap();
+        let existing = UserToken {
+            name: "alice".to_string(),
+            spn: "oidc:sub:old-subject".to_string(),
+            uuid: object_id,
+            real_gidnumber: Some(1000),
+            gidnumber: 1000,
+            displayname: "Alice".to_string(),
+            shell: None,
+            groups: vec![],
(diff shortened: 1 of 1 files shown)


dmulder:
I'm putting this one on the back burner, since I think it needs more discussion and careful thought first.

PR #1569, opened 2026-08-03
Add aad-tool kinit command

### LINKED ISSUE #1568: Get kerberos tokens for other accounts
### Summary

Let users get kerberos tickets for another user that is not the one that is logged in.

### Problem Statement

Currently, we get a kerberos token for the logged in account. However, some companies require one account for basic usage and one for administrative things. So, I can be logged in with my basic user and have access to an azure files share with my administrative account. With Himmelblau as is, I have no way to mount this share on my computer without logging in with my admin account (which might not be allowed).

With "normal" kerberos, I can get a token for my admin account, but not with Himmelblau AFAIK.

### Proposed Solution

I'm thinking something that mimics `kinit <user>`, just that it works with Himmelblau.

Alternatively, a way to register a second account that can't be logged in to but gets kerberos tokens when logging in with the basic user.

### Benefits

Allow user's to access file shares with Kerberos without changing their logged in account.

### Alternatives Considered

Logging in with the second account on the computer and accessing the share from that.

### Additional Context

_No response_

### Code of Conduct

---

### 💡 Help Make It Happen!

For US tax exempt donations:

himmelbot:
## Summary
- Add `aad-tool kinit` for initializing Kerberos tickets for a named Entra ID account.
- Authenticate the target account without running normal login ccache side effects, then ask himmelblaud to store the target account's tickets in the calling user's credential cache.
- Parse MIT kinit-compatible argument names where practical and fail explicitly for unsupported options.

Fixes #1568

## Testing
- `rustfmt --edition 2021 src/common/src/unix_proto.rs src/daemon/src/daemon.rs src/cli/src/main.rs src/cli/src/opt/tool.rs`
- `git diff --check`
- `cargo check -p aad-tool -p himmelblaud` (blocked: this environment lacks pkg-config/OpenSSL/dbus development discovery)

### DIFF
--- a/src/cli/src/main.rs
+++ b/src/cli/src/main.rs
@@ -38,9 +38,13 @@ use broker_client::BrokerClient;
 use clap::Parser;
 use himmelblau::{error::MsalError, graph::Graph, AuthOption, BrokerClientApplication};
 use himmelblau::{ConfidentialClientApplication, UserToken};
-use himmelblau_unix_common::auth::{authenticate_async, SimpleMessagePrinter};
+use himmelblau_unix_common::auth::{
+    authenticate_async, authenticate_with_client_for_reuse, wait_for_daemon_client, MessagePrinter,
+    SimpleMessagePrinter,
+};
 use himmelblau_unix_common::auth_handle_mfa_resp;
 use himmelblau_unix_common::client::call_daemon;
+use himmelblau_unix_common::client_sync::DaemonClientBlocking;
 use himmelblau_unix_common::config::{parse_ttl_to_seconds, split_username, HimmelblauConfig};
 use himmelblau_unix_common::constants::{
     CONFIDENTIAL_CLIENT_CERT_KEY_TAG, CONFIDENTIAL_CLIENT_CERT_TAG, CONFIDENTIAL_CLIENT_SECRET_TAG,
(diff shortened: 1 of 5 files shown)


src/cli/src/main.rs
```diff
@@ -1644,36 +1710,112 @@ async fn main() -> ExitCode {
...
-            )
-            .await
-            {
-                PamResultCode::PAM_SUCCESS => return ExitCode::SUCCESS,
-                e => {
-                    error!("Authentication failed: {:?}", e);
-                    return ExitCode::FAILURE;
-                }
-            }
+            authenticate_account(cfg, principal, opts).await
```
dmulder: This entirely misses the point. This will authenticate a user, causing himmelblaud to store _that_ users ticket in _their own_ ccache. This does _not_ accomplish the stated goal of authenticating as a _different_ user, and storing that _other_ user's ticket in their ccache.
For example, we should be able to authenticate to the host as `bob`, then run `aad-tool kinit admin`, authenticate via `aad-tool` as `admin`, and have the `admin` ticket stored in `bob`s ccache.
This will require obtaining the raw PRT for `admin` via libhimmelblau auth, then using the ccache logic in src/daemon/src/tasks_daemon.rs (some of it duplicated here) to store the `admin` ticket in `bob`s ccache.

dmulder (changes requested):

dmulder:
@copilot resolve the merge conflicts in this pull request

PR #1603, opened 2026-08-12
Bound custom compliance script execution and stop mokutil hanging himmelblaud-tasks

### LINKED ISSUE #1602: himmelblaud-tasks is SIGABRT'd by its own watchdog when an Intune CustomCompliance script hangs
## Summary

`himmelblaud_tasks` executes Intune CustomCompliance discovery scripts with a
blocking, **unbounded** `std::process::Command::output()`. Any script that hangs
parks the thread forever, the service stops pinging the systemd watchdog, and
`WatchdogSec=120s` SIGABRTs the whole task daemon.

The compliance check then never completes, so Intune never receives updated
results and silently retains the **previous** verdict indefinitely. On this
machine that meant the Intune portal kept reporting "Secure Boot is not enabled"
for two days after Secure Boot had actually been enabled, with a
"device will lose access to company resources" warning.

There is a concrete, reproducible trigger for the hang (details below), but the
report here is about the missing timeout: **a badly-behaved tenant script should
not be able to kill the task daemon.**

## Environment

| | |
|---|---|
| himmelblau | 4.0.0-1.x86_64 (`main` @ `3a8c96e`) |
| OS | Fedora Linux 44 (Workstation) |
| Kernel | 7.1.8-200.fc44.x86_64 |
| mokutil | 0.7.2-3.fc44 |
| Config | `apply_policy = true`, `enable_experimental_intune_custom_compliance = true` |

## Observed behaviour

```
himmelblaud-tasks.service: Watchdog timeout (limit 2min)!
himmelblaud-tasks.service: Killing process 2840 (himmelblaud_tas) with signal SIGABRT.
himmelblaud-tasks.service: Killing process 12781 (bash) with signal SIGABRT.
himmelblaud-tasks.service: Killing process 12803 (mokutil) with signal SIGABRT.
himmelblaud-tasks.service: Killing process 12804 (grep) with signal SIGABRT.
himmelblaud-tasks.service: Main process exited, code=dumped, status=6/ABRT
himmelblaud-tasks.service: Failed with result 'watchdog'.
```

and correspondingly in `himmelblaud`:

```
WARN  compliance check: task error: channel closed
WARN  Intune failure: channel closed
```

`aad-tool compliance-check` reports only:

```
policy application failed: compliance check could not complete.
```

This repeated on every policy cycle and every manual `aad-tool compliance-check`
invocation. The debug line `Custom compliance script execution finished` was
never reached.

## Root cause

`src/policies/src/custom_compliance_ext.rs:66-73`:

```rust
let output = if has_shebang {
    Command::new(&temp_path)
        .output()
        .context("Failed to execute script directly")?
} else {
    Command::new("/bin/sh")
        .arg(&temp_path)
        .output()
        .context("Failed to execute script via /bin/sh")?
};
```

`std::process::Command::output()` blocks in `wait()` with no timeout. Because
`execute_script()` is a synchronous function called from the async
`apply_compliance()` (line 198), the hang parks a Tokio worker thread and the
watchdog ticker never runs.

`WatchdogSec=120s` was added to this unit later, in #1266, so this interaction
appears never to have been exercised.

## Suggested fix

Bound the script execution. Two things matter:

1. **A timeout.** 60s or so, ideally configurable.
2. **Kill the whole process group, not just the direct child.** The discovery
   script is `bash`, and the process that hangs is its *grandchild* (`mokutil`).
   Killing only the direct child leaves an orphaned process spinning at 100% CPU
   forever. Use `CommandExt::process_group(0)` and `killpg()`.
3. **Never block on draining the output pipes.** This one is easy to get wrong:
   `read_to_end()` on a child's stdout only returns once *every* writer has
   closed it, and a process that escaped the group (`setsid()`, or any daemon
   the script backgrounds) holds the inherited write end open for as long as it
   lives. Waiting on the reader threads therefore reintroduces the exact hang
   the timeout was added to prevent — including on the *success* path, where the
   script itself exited cleanly. The readers must be abandoned on timeout, and
   bounded (not `join()`ed) after a normal exit.

Implementing 1 and 2 without 3 still hangs; we hit this while testing.

On timeout, failing the check with a clear error is strictly better than the
current behaviour — the daemon survives, and the operator gets an actionable log
line instead of `channel closed`.

It would also be worth logging the script's runtime at debug level; the current
logs give no indication that a script is the thing that stalled.

## Related: the unit's own sandbox triggers this

The specific hang was a three-layer interaction, filed separately upstream:

1. **kernel** — `efi_mokvar_sysfs_read()` in `drivers/firmware/efi/mokvar-table.c`
   returns `0` (not `-EPERM`) when the caller lacks `CAP_SYS_ADMIN`, which is
   indistinguishable from EOF.
2. **mokutil** — `mok_get_variable()` in `src/util.c` has a
   `while (pos < bufsz) { ssz = read(...); ... pos += ssz; }` loop with **no
   `ssz == 0` guard**, so it spins forever at 100% CPU.
3. **himmelblau** — this issue.

`himmelblaud-tasks.service` runs `User=root` with a `CapabilityBoundingSet` that
excludes `CAP_SYS_ADMIN`. That combination is exactly what triggers layers 1+2:

| context | `open()` on `/sys/firmware/efi/mok-variables/db` | result |
|---|---|---|
| non-root | fails `EACCES` (files are mode `0400`) | falls back to efivarfs, works, ~0.04s |
| root + `CAP_SYS_ADMIN` | succeeds, `read()` returns data | works |
| **root without `CAP_SYS_ADMIN`** | succeeds, `read()` returns `0` forever | **spins at 100% CPU** |

So the daemon's own hardening puts `mokutil` into the one state that hangs, and
Microsoft's stock Secure Boot discovery script for Linux calls `mokutil --db`.
Any tenant using it on a distro where himmelblau runs the task daemon as root
with reduced capabilities will hit this.

**Suggested additional fix** — add to `himmelblaud-tasks.service` in
`scripts/gen_servicefiles.py`:

```ini
InaccessiblePaths=/sys/firmware/efi/mok-variables
```

This makes `mokutil`'s `open()` fail so it takes its normal efivarfs fallback —
the same path unprivileged callers already use. Verified working here under the
unit's full hardening set. It is preferable to granting `CAP_SYS_ADMIN`, which
would be a near-root hardening regression, and it does not affect
`mokutil --sb-state` (that uses `efi_get_variable()` and never touches
`mok-variables`).

## Verification

With `InaccessiblePaths=/sys/firmware/efi/mok-variables` added as a drop-in:

```
Custom compliance script execution finished | exit_code: Some(0) | stdout_len: 137
CustomComplianceCSE: script executed successfully | output:
  {"SecureBootEnabled":"true","SecureBootCA2023":"Present",
   "HardwareExcluded":"false","ExcludeReason":"none","ScriptVersion":"2026.07.01"}
Intune compliance status | device_info.compliance_state: Compliant
```

`aad-tool compliance-check` → `compliance check passed`, `NRestarts=0`, no
further watchdog kills.

Note that this only fixes the *trigger*. The missing timeout in
`execute_script()` remains, and any other slow or hung tenant script will still
take down the task daemon.

---

### 💡 Help Make It Happen!

For US tax exempt donations:

glima:
## Summary

Fixes #1602

`himmelblaud_tasks` runs Intune CustomCompliance discovery scripts with an
unbounded `Command::output()`. A script that never returns parks a Tokio worker,
the service stops pinging its systemd watchdog, and `WatchdogSec=120s` SIGABRTs
the whole task daemon — taking Kerberos, policy application and compliance
reporting with it. The only symptom is `compliance check: task error: channel closed`.

Because the check never completes, Intune silently retains the *previous*
verdict forever. Here that meant the portal kept reporting "Secure Boot is not
enabled" for two days after Secure Boot was actually enabled, with a
"device will lose access to company resources" warning.

## What Changed

Two commits, each independently useful:

**1. `systemd: hide mok-variables from himmelblaud-tasks`** — removes the trigger.

Microsoft's stock Secure Boot compliance script runs `mokutil --db`, which under
this unit never returns. The cause spans three layers:

1. `efi_mokvar_sysfs_read()` returns `0`, not `-EPERM`, when the reader lacks
   `CAP_SYS_ADMIN` — indistinguishable from EOF.
2. mokutil's `mok_get_variable()` read loop has no `ssz == 0` guard, so it spins
   at 100% CPU forever.
3. `himmelblaud-tasks` runs `User=root` with a `CapabilityBoundingSet` that
   excludes `CAP_SYS_ADMIN` — exactly the state that triggers 1 and 2.

Counter-intuitively, *less* privilege works. The files are mode `0400`, so an
unprivileged caller's `open()` fails `EACCES` and mokutil takes its efivarfs
fallback. Only *partial* privilege hangs:

| context | `open()` on `mok-variables/db` | result |
|---|---|---|
| non-root | fails `EACCES` | efivarfs fallback, works, ~0.04s |
| root + `CAP_SYS_ADMIN` | succeeds, returns data | works |
| **root without `CAP_SYS_ADMIN`** | succeeds, `read()` returns `0` forever | **spins at 100% CPU** |

`InaccessiblePaths` forces that same fallback. Preferred over granting
`CAP_SYS_ADMIN`, which would be a near-root hardening regression. It does not
affect `mokutil --sb-state`, which uses `efi_get_variable()` and never touches
`mok-variables`. Gated on systemd >= 231 via the existing `MINVER` mechanism;
the `-` prefix makes it a no-op on non-EFI systems.

Filed upstream separately for layers 1 and 2 (kernel, mokutil).

**2. `policies: bound custom compliance script execution`** — fixes the class of bug.

These scripts are tenant-supplied and arbitrary; a hostile or merely buggy one
should not be able to kill the daemon. Adds `output_with_timeout()` with a 60s
budget (under `WatchdogSec`, with headroom). Three details matter:

- **Signal the process *group*.** The script is a shell and the process that
  hangs is a grandchild; killing only the child leaves it orphaned and spinning.
- **Don't wait on the pipe readers after the group kill.** `read_to_end()`
  returns only once every writer closes the pipe, and a process that escaped via
  `setsid()` holds the inherited write end open for as long as it lives —
  reintroducing the exact hang being fixed.
- **Bound the drain on the success path too**, for the same reason: a script can
  exit cleanly having backgrounded a process that still holds the pipe.

Implementing the first two without the third still hangs; we hit that while testing.

Also logs the script's runtime, which previously gave no hint that a script was
the thing that stalled.

`Cargo.nix` is regenerated for the new `libc` dependency (hand-applied in the
generated form; `crate2nix` wasn't available on this host — worth a re-run by
someone with nix).

## Testing

- **Distro(s) tested:** Fedora Linux 44 (Workstation), kernel 7.1.8-200.fc44,
  mokutil 0.7.2-3.fc44, himmelblau 4.0.0-1. Test suite additionally exercised
  under Debian/dash in the `make test` container.
- **Steps performed:**
  1. Reproduced deterministically with `aad-tool compliance-check` — watchdog
     SIGABRT every cycle, `NRestarts` climbing.
  2. Caught `mokutil` live: `State: R`, 99.7% CPU — a busy loop, not blocked I/O.
  3. Bisected the unit's hardening one directive at a time via `systemd-run`;
     only `CapabilityBoundingSet` reproduced it.
  4. Binary-searched the 35 capabilities absent from the unit's set → `CAP_SYS_ADMIN`.
  5. Confirmed `InaccessiblePaths` fixes rule 2 (`mokutil --db`) under the unit's
     full hardening set, and that rule 1 (`mokutil --sb-state`) still works.
  6. Applied as a drop-in on the live host and re-ran the real compliance check.
  7. Built `make fedora44` and confirmed the directive is present in the shipped
     `himmelblaud-tasks.service` inside the RPM.
- **Results observed:**
  - Before: `Watchdog timeout (limit 2min)!` → `status=6/ABRT`, script never finished.
  - After: `Custom compliance script execution finished | exit_code: Some(0)`,
    `{"SecureBootEnabled":"true","SecureBootCA2023":"Present",...}`,
    `Intune compliance status | device_info.compliance_state: Compliant`,
    `aad-tool compliance-check` → `compliance check passed`, `NRestarts=0`,
    no further watchdog kills. Intune portal now reports the device compliant.

Each of the six new unit tests was mutation-tested — I reverted each fix
individually and confirmed the corresponding test fails, and fails *fast*
rather than hanging:

| test | regression it pins |
|---|---|
| `passes_through_status_and_streams` | exit status / stdout / stderr still propagate |
| `large_output_does_not_deadlock` | >64KB output doesn't deadlock on the pipe buffer |
| `hanging_script_is_killed_at_timeout` | the timeout fires at all |
| `timeout_kills_grandchildren` | `killpg` vs. `child.kill()` — orphan left spinning |
| `timeout_returns_even_if_a_process_escapes_the_group` | no `join()` on the timeout path |
| `exited_script_leaving_a_pipe_holder_does_not_hang` | bounded drain on the success path |

## Automated Checks

- [x] I ran `make test` (full suite green)
- [x] I ran the distro package build target (`make fedora44`; verified the
      generated unit inside the resulting RPM)
- [x] I ran `make test-selinux` (if applicable) — **N/A**, no SELinux policy
      files touched

## System Integration Impact

Touches systemd. `himmelblaud-tasks.service` gains one directive:

```ini
InaccessiblePaths=-/sys/firmware/efi/mok-variables
```

- Emitted only for systemd >= 231; older systemd is unaffected.
- The `-` prefix makes it a no-op where the path doesn't exist (non-EFI, BIOS,
  containers, systems without shim).
- Narrows the sandbox rather than widening it — no new privileges are granted.
- Only affects processes started by `himmelblaud-tasks`. `mokutil` invoked from
  anywhere else is untouched.
- Behaviour change for compliance scripts: `mokutil --db` now returns via the
  efivarfs path instead of hanging. `--sb-state` is unaffected.
- No PAM/NSS/authselect, SELinux/AppArmor, filesystem path or credential changes.
- New runtime dependency for `himmelblau_policies`: `libc` (already in the
  workspace), reflected in `Cargo.lock`/`Cargo.nix`.

## Checklist

- [x] I manually tested this change on a test VM — tested on real hardware
      (Fedora 44 laptop) rather than a VM; the bug is firmware/EFI-dependent and
      does not reproduce in a VM without matching Secure Boot + MOK state.
- [x] PR scope is focused and linked to an issue/enhancement/discussion (#1602)

Happy to split this into two PRs if you'd prefer the systemd change and the
timeout hardening reviewed separately — they're independent commits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2265,6 +2265,7 @@ dependencies = [
  "async-trait",
  "base64 0.23.1",
  "himmelblau_unix_common",
+ "libc",
  "libhimmelblau",
  "os-release",
  "regex",
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -7440,6 +7440,10 @@ rec {
             packageId = "himmelblau_unix_common";
           }
           {
+            name = "libc";
+            packageId = "libc";
+          }
+          {
             name = "libhimmelblau";
             packageId = "libhimmelblau";
             features = [ "broker" "changepassword" "on_behalf_of" "mfa_method_selection" "optional_mfa" "intune_portal_vers_selection" "redirect_uri" "pop_support" "ipvers" "set_timeout" ];
--- a/scripts/gen_servicefiles.py
+++ b/scripts/gen_servicefiles.py
@@ -40,6 +40,7 @@ MINVER = {
     "ProtectKernelLogs": 244,
     "ProtectControlGroups": 232,
     "MemoryDenyWriteExecute": 231,
+    "InaccessiblePaths": 231,
     "CacheRuntimeStateDirs": 235,  # CacheDirectory/RuntimeDirectory/StateDirectory
     "ConditionPathExists": 12,
     "LoadCredentialEncrypted": 250,
@@ -266,6 +267,27 @@ WantedBy=multi-user.target
     rw_paths = "/home /run/himmelblaud /tmp /etc/krb5.conf.d /etc /var/lib /var/cache/nss-himmelblau /var/cache/himmelblau-policies"
     rw_line = f"ReadWritePaths={rw_paths}" if rw_paths_available else ""
 
+    # Intune CustomCompliance discovery scripts commonly shell out to `mokutil`
+    # (Microsoft's stock Secure Boot compliance script runs `mokutil --db`).
+    #
+    # mokutil's mok_get_variable() reads /sys/firmware/efi/mok-variables/<name>
+    # first and only falls back to efivarfs if that open() fails. Those files are
+    # mode 0400 root:root, so this unit (User=root) opens them successfully - but
+    # the kernel's efi_mokvar_sysfs_read() returns 0 rather than -EPERM when the
+    # reader lacks CAP_SYS_ADMIN, which our CapabilityBoundingSet drops. mokutil's
+    # read loop has no ssz==0 guard, so it spins at 100% CPU forever, blocks the
+    # task daemon's watchdog pings, and WatchdogSec SIGABRTs the whole service.
+    #
+    # Hiding the directory makes that open() fail so mokutil takes its normal
+    # efivarfs fallback - the same path unprivileged callers already use. This is
+    # preferable to granting CAP_SYS_ADMIN, and does not affect `mokutil
+    # --sb-state`, which uses efi_get_variable() and never touches mok-variables.
+    tasks_inaccessible = (
+        "InaccessiblePaths=-/sys/firmware/efi/mok-variables"
+        if supported("InaccessiblePaths")
+        else ""
+    )
+
     tasks_unit = f"""\
 # You should not need to edit this file. Instead, use a drop-in file:
 #   systemctl edit himmelblaud-tasks.service
@@ -299,6 +321,7 @@ AmbientCapabilities=CAP_SETUID CAP_SETGID
 {rw_line}
 {os.linesep.join([h for h in tasks_hardening if not h.startswith('ProtectSystem=')])}
 {tasks_private_devices}
+{tasks_inaccessible}
 
 [Install]
 WantedBy=multi-user.target
--- a/src/policies/Cargo.toml
+++ b/src/policies/Cargo.toml
@@ -27,6 +27,7 @@ os-release = "0.1.0"
 semver = "1.0.28"
 libhimmelblau.workspace = true
 tempfile = "3.27.0"
+libc = { workspace = true }
 
 [package.metadata.cargo-machete]
 ignored = ["libhimmelblau"]
--- a/src/policies/src/custom_compliance_ext.rs
+++ b/src/policies/src/custom_compliance_ext.rs
@@ -23,11 +23,268 @@ use base64::Engine;
 use himmelblau::intune::{ComplianceState, IntuneStatus, PolicyStatus};
 use himmelblau_unix_common::config::HimmelblauConfig;
 use std::fs::Permissions;
-use std::io::Write;
+use std::io::{self, Write};
 use std::os::unix::fs::PermissionsExt;
-use std::process::Command;
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::os::unix::process::CommandExt;
+use std::process::{ChildStderr, ChildStdout, Command, Output, Stdio};
+use std::time::{Duration, Instant};
 use tempfile::{NamedTempFile, TempPath};
 use tracing::{debug, error};
+
+/// Maximum wall-clock time a tenant-supplied discovery script may run for.
+const SCRIPT_TIMEOUT: Duration = Duration::from_secs(60);
+
+/// Minimum extra time allowed for the pipes to drain after the script has
+/// exited. Only whatever is still sitting in the pipe buffers remains at that
+/// point, so this is generous; it exists so that a script finishing right on
+/// the deadline does not lose its output.
+const READER_GRACE: Duration = Duration::from_secs(5);
+
+/// How long a single `poll()` waits before the loop re-checks the child and the
+/// deadline. Bounds how late a timeout can fire, nothing else.
+const POLL_INTERVAL_MS: libc::c_int = 50;
+
+/// Upper bound on how much a script may write to each stream. A discovery
+/// script is expected to emit a small JSON object; anything beyond this is
+/// runaway output and must not be allowed to grow the daemon's heap without
+/// limit for the whole timeout budget.
+const MAX_STREAM_BYTES: usize = 16 * 1024 * 1024;
+
+/// A child pipe read without blocking.
+///
+/// Reader *threads* are deliberately avoided. A thread parked in `read_to_end()`
+/// cannot be cancelled and only returns once every writer has closed the pipe -
+/// which a process that escaped the child's process group (e.g. via `setsid()`)
+/// can prevent for as long as it chooses to live. Abandoning such threads leaks
+/// a thread, its stack and an unbounded buffer per invocation, so instead the
+/// descriptors are drained non-blockingly from the one loop that owns them and
+/// closed when the budget expires.
+struct Pipe {
+    /// `None` once EOF (or an error) has been seen, or once it is dropped.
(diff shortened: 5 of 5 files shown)


dmulder (approved):
LGTM

PR #1620, opened 2026-08-17
Skip Hello TOTP for local password-only sessions

### LINKED ISSUE #1413: Allow skipping Hello TOTP on local auth
### Summary

Hello TOTP is a nice alternative to the MS MFA for remote login.

However, it would be nice if it could be skipped for local authentication.

I would have assumed from the description `Allow password-only auth for local console sessions.` that the option `allow_console_password_only` would do that, but apparently it does not.

So right now, the TOTP code is prompted for on each authentication attempt, both local and remote.

### Problem Statement

Grabbing my secondary device each time I authenticate locally is not great, and I would assume that was the intention behind the `allow_console_password_only` option.

### Proposed Solution

Make `allow_console_password_only` affect the Hello TOTP as it does for MFA, or add some new option that allows skipping the TOTP code prompt when authenticating locally.

### Benefits

Makes Hello TOTP a more direct replacement for MFA.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Code of Conduct

---

### 💡 Help Make It Happen!

[Donate to Our Collective](https://himmelblau-idm.org/backers.html#backers-)

For US tax exempt donations:

himmelbot:
## Summary
- Skip Hello TOTP setup/prompt after successful Hello PIN auth when the session is local and `allow_console_password_only` is enabled.
- Preserve existing Hello TOTP behavior for remote sessions and for local sessions where console password-only auth is disabled.

Fixes #1413.

## Testing
- `git diff --check`
- Not run: `cargo fmt --check` / Rust checks because `cargo` is not installed in this environment.

### DIFF
--- a/src/common/src/idprovider/himmelblau.rs
+++ b/src/common/src/idprovider/himmelblau.rs
@@ -1826,6 +1826,7 @@ impl IdProvider for HimmelblauProvider {
                 .iter()
                 .any(|s| !s.is_empty() && service.contains(s));
         let hello_totp_enabled = check_hello_totp_enabled!(self);
+        let console_password_only = self.config.lock().await.get_allow_console_password_only();
         let allow_remote_hello = self.config.lock().await.get_allow_remote_hello();
         // Skip Hello authentication if it is disabled by config
         let hello_enabled = self.config.lock().await.get_enable_hello();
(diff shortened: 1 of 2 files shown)


dmulder (dismissed):
LGTM

PR #1626, opened 2026-08-18
fix(profile): preserve desktop session selection

bash-bandicoot:
## Summary

| Section | Detail |
|---|---|
| TL;DR | Profile-photo updates now use AccountsService, preserving the desktop selected in GDM. |
| Why | Truncating the user record removed `XSession=i3` and made GDM fall back to GNOME. |
| Risk | 4/10 because D-Bus and SELinux behavior spans several Linux distributions |
| Confidence | 9/10 because local validation and GitHub checks passed |

<details>
<summary>Test evidence</summary>

| Test | Before | After |
|---|---|---|
| Preserve desktop selection | The photo task truncated the user record, deleting `Session` and `XSession`. | `SetIconFile` updates the icon while AccountsService retains other fields. |
| Publish complete photos | The canonical icon was truncated before download completed. | AccountsService publishes a private temporary file only after download succeeds. |
| Daemon tests | - | Focused task tests: 8 passed |
| Format and metadata | - | Format, locked metadata, and diff checks passed |
| SELinux policy | - | Passed Rocky 8/9/10, Fedora 43/44, and Tumbleweed; SLE 16 needs SCC credentials |
| Clippy | - | Blocked by pre-existing unknown lint `clippy::manual_checked_ops` |

</details>

## What Changed

- Update icons through AccountsService D-Bus instead of rewriting user records.
- Download to a private temporary file before publishing.
- Permit tasks-daemon D-Bus messages with `accountsd_t` under SELinux.

## Testing

- **Distro(s) tested:** Rocky 8/9/10, Fedora 43/44, and Tumbleweed policy containers.
- **Steps performed:** Focused daemon tests, format/metadata checks, and SELinux policy harness.
- **Results observed:** 8 daemon tests and six policy targets passed; SLE 16 needs SCC credentials.

## Automated Checks

- [x] Focused daemon tests and SELinux policy checks passed.
- [ ] Full `make test` and package builds were not run.

## System Integration Impact

Adds an AccountsService D-Bus dependency and narrow SELinux messaging rules for `accountsd_t`.

## Checklist

- [ ] No disposable desktop VM was available for manual GDM testing.
- [ ] No public issue exists; the regression was reported internally.

### DIFF
--- a/src/daemon/Cargo.toml
+++ b/src/daemon/Cargo.toml
@@ -51,9 +51,8 @@ libhimmelblau.workspace = true
 url.workspace = true
 console-subscriber = { workspace = true, optional = true }
 libkrimes = { workspace = true }
-
-[dev-dependencies]
 tempfile = "3.27.0"
+zbus = { workspace = true }
 
 [package.metadata.deb]
 name = "himmelblau"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2358,6 +2358,7 @@ dependencies = [
  "uuid",
  "uzers",
  "walkdir",
+ "zbus",
 ]
 
 [[package]]
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -7793,6 +7793,10 @@ rec {
             packageId = "sketching";
           }
           {
+            name = "tempfile";
+            packageId = "tempfile";
+          }
+          {
             name = "tokio";
             packageId = "tokio";
             features = [ "rt" "macros" "sync" "time" "net" "io-util" "signal" "rt-multi-thread" "tracing" ];
@@ -7823,11 +7827,9 @@ rec {
             name = "walkdir";
             packageId = "walkdir";
           }
-        ];
-        devDependencies = [
           {
-            name = "tempfile";
-            packageId = "tempfile";
+            name = "zbus";
+            packageId = "zbus";
           }
         ];
         features = {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -143,6 +143,7 @@ reqwest = { version = "^0.13.1", features = ["json"] }
 anyhow = "^1.0.102"
 tokio = { version = "^1.52.2", features = ["rt", "macros", "sync", "time", "net", "io-util", "signal", "rt-multi-thread"] }
 tokio-util = { version = "^0.7.18", features = ["codec"] }
+zbus = "5.14"
 async-trait = "^0.1.89"
 himmelblau_policies = { path = "src/policies" }
 pem = "^3.0.6"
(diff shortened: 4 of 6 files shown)


bash-bandicoot:
Addressed: the description now includes all required template sections and is 2,255 bytes, below the 2,500-character limit.

bash-bandicoot:
@dmulder and/or @bluca , please take a look when you have time

dmulder:
> Addressed: the description now includes all required template sections and is 2,255 bytes, below the 2,500-character limit.

You really don't need to worry about these. If it didn't cause the PR to close, then just ignore it.
Those checks are to watch for bots that submit large meaningless PRs. If the bots hit too many of the thresholds, it gets closed without review.

dmulder:
> You really don't need to worry about these. If it didn't cause the PR to close, then just ignore it. Those checks are to watch for bots that submit large meaningless PRs. If the bots hit too many of the thresholds, it gets closed without review.

Oh! I see it did close it.

src/daemon/Cargo.toml
```diff
@@ -51,9 +51,8 @@ libhimmelblau.workspace = true
 url.workspace = true
 console-subscriber = { workspace = true, optional = true }
 libkrimes = { workspace = true }
-
-[dev-dependencies]
 tempfile = "3.27.0"
+zbus = "5.14"
```
dmulder: Shouldn't this be `zbus = { workspace = true }`, so that we mirror the zbus workspace version, instead of defining it explicitly to version 5.14?

dmulder (changes requested):

dmulder:
The CI failure is because you neglected to install the git hooks before committing (`make setup-hooks`). You can fix it by manually running `make crate2nix-generate`.

src/daemon/Cargo.toml
bash-bandicoot: Fixed, thanks. I added `zbus` to `workspace.dependencies` and changed this to `zbus = { workspace = true }`.

bash-bandicoot:
> The CI failure is because you neglected to install the git hooks before committing (`make setup-hooks`). You can fix it by manually running `make crate2nix-generate`.

Fixed. I ran `make setup-hooks` locally and regenerated `Cargo.nix` with `make crate2nix-generate`. The updated generated file is included in the latest push.

src/daemon/Cargo.toml
```diff
@@ -54,6 +54,7 @@ libkrimes = { workspace = true }
 
 [dev-dependencies]
 tempfile = "3.27.0"
+zbus = { workspace = true }
```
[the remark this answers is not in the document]
bash-bandicoot: Good catch. `tempfile` and `zbus` are used by the `himmelblaud_tasks` binary, so I moved them back to `[dependencies]` while keeping `zbus = { workspace = true }`, then regenerated `Cargo.nix`.

I also added debug logging for `FindUserByName` failures before falling back to `CacheUser`.

PR #1651, opened 2026-08-27
Restrict SYSTEMD_ACTIVATION_UNIT bypass to authenticated systemd-executor

dmulder:
### Motivation

- A previous change let `should_skip_daemon_call()` unconditionally trust the `SYSTEMD_ACTIVATION_UNIT` environment variable, allowing untrusted callers to force PAM/NSS hooks to skip daemon checks.
- The intent is to avoid deadlocks during legitimate `sd-executor` service startup while preventing an attacker-controlled environment from bypassing authentication and account checks.

### Description

- Tighten `should_skip_daemon_call()` in `src/common/src/client_sync.rs` to only return true when the process appears to be the trusted systemd executor by checking `geteuid() == 0`, parent PID is `1`, and `/proc/self/exe` basename equals `systemd-executor`, in addition to matching the Himmelblau unit names.
- Preserve the existing unit-name check for `himmelblaud.service` and `himmelblaud-tasks.service` and keep the `OnceLock` caching semantics.
- No changes were made to PAM/NSS call sites; they continue to call `should_skip_daemon_call()` as before.

### Testing

- Ran `cargo fmt --all -- --check`, which succeeded.
- Ran repository diff and formatting checks (`git diff HEAD^ --check`), which succeeded.
- Ran `cargo test -p himmelblau_unix_common --lib`, which failed to complete due to a missing system dependency (`dbus-1` development library required by `libdbus-sys`), so full test-suite execution could not be completed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a76189523648326899f5f819436fac1)

### DIFF
--- a/src/common/src/client_sync.rs
+++ b/src/common/src/client_sync.rs
@@ -30,10 +30,20 @@ pub fn should_skip_daemon_call() -> bool {
 
     static SKIP: OnceLock<bool> = OnceLock::new();
     *SKIP.get_or_init(|| {
-        matches!(
+        // SYSTEMD_ACTIVATION_UNIT may be inherited from an untrusted caller.
+        // Only sd-executor, running directly under the system manager, may use
+        // it to suppress daemon calls during service credential resolution.
+        let is_root_system_child = unsafe { libc::geteuid() == 0 && libc::getppid() == 1 };
+        let is_systemd_executor = std::fs::read_link("/proc/self/exe")
+            .ok()
+            .and_then(|path| path.file_name().map(|name| name == "systemd-executor"))
+            .unwrap_or(false);
+        let is_himmelblau_unit = matches!(
             std::env::var_os("SYSTEMD_ACTIVATION_UNIT").as_deref(),
             Some(v) if v == "himmelblaud.service" || v == "himmelblaud-tasks.service"
-        )
+        );
+
+        is_root_system_child && is_systemd_executor && is_himmelblau_unit
     })
 }
 


bluca:
This doesn't look right to me, the executor binary doesn't even exist on older versions, and even in new versions it might be disabled if built with `systemd-multicall-binary=true`. It's an internal implementation detail, not an API, and might go away at any given time, which means this will break.

dmulder:
> This doesn't look right to me, the executor binary doesn't even exist on older versions, and even in new versions it might be disabled if built with `systemd-multicall-binary=true`. It's an internal implementation detail, not an API, and might go away at any given time, which means this will break.

Sorru @bluca, I didn't see your comment before merging this.

dmulder:
@bluca perhaps we should instead just include the root child check? If SYSTEMD_ACTIVATION_UNIT is set, and the calling process is a root child, I think we could safely assume it wasn't called by a malicious actor.

bluca:
> > This doesn't look right to me, the executor binary doesn't even exist on older versions, and even in new versions it might be disabled if built with `systemd-multicall-binary=true`. It's an internal implementation detail, not an API, and might go away at any given time, which means this will break.
> 
> Sorru @bluca, I didn't see your comment before merging this.

Would have been impressive if you did, given I wrote it _after_ it was already merged :-P

> @bluca perhaps we should instead just include the root child check? If SYSTEMD_ACTIVATION_UNIT is set, and the calling process is a root child, I think we could safely assume it wasn't called by a malicious actor.

Having that one check seems harmless, sure. But I'm not entirely sure what it is supposed to guard against?

dmulder:
> Would have been impressive if you did, given I wrote it _after_ it was already merged :-P

Lol, I thought I had just carelessly merged and missed your comment.

> Having that one check seems harmless, sure. But I'm not entirely sure what it is supposed to guard against?

It was flagged by codex security, since a malicious user could set `SYSTEMD_ACTIVATION_UNIT` and effectively disable Himmelblau and fallback to local auth (it forces the pam module to return PAM_IGNORE). Assuming the system is configured correctly (pam_unix doesn't successfully authenticate a Himmelblau user), then this is effectively just a DOS on yourself. Not a huge issue, really.

bluca:
> It was flagged by codex security, since a malicious user could set `SYSTEMD_ACTIVATION_UNIT` and effectively disable Himmelblau and fallback to local auth (it forces the pam module to return PAM_IGNORE). Assuming the system is configured correctly (pam_unix doesn't successfully authenticate a Himmelblau user), then this is effectively just a DOS on yourself. Not a huge issue, really.

But how would they do that? If they have complete control of the pam setup, then they can just swap the config or modules away

dmulder:
I'm saying a regular user with no privileges could do `SYSTEMD_ACTIVATION_UNIT=1 su - tux@himmelblau-idm.org` and try to authenticate the user via pam_unix, skipping pam_himmelblau.

Naturally, if pam_unix fails to authenticate the user, it will still fail. I don't really see this as a security issue, honestly. A poorly configured host _might_ encounter this issue though.

bluca:
Sure but such a user could also run the same in a network namespace with no interfaces, forcing the offline authentication path, no?

dmulder:
> Sure but such a user could also run the same in a network namespace with no interfaces, forcing the offline authentication path, no?

I suppose so. That's an entirely different issue, but likewise sounds incredibly unlikely to be successful. I think what we're really trying to avoid here is lots of mini issues being chained together into a successful attack. Perhaps an attacker with AI could chain the pam_unix downgrade with some flaw in pam_unix?

PR #1655, opened 2026-08-27
Delay breakglass password caching until MFA succeeds

dmulder:
### Motivation

- The code cached an offline password verifier as soon as the primary password validated while an MFA challenge was still in progress, allowing an attacker who abandons MFA to later authenticate offline without the second factor.
- The change aims to restore the previous protection by ensuring password verifiers are only persisted after a completed successful authentication, preventing an MFA downgrade to SFA when offline breakglass is enabled.

### Description

- Removed requests to update the offline password cache from all MFA challenge branches so MFA challenge responses now return `AuthCacheAction::None` until MFA completes. (changed MFA handling in `src/common/src/idprovider/himmelblau.rs`).
- Moved the logic that may request `AuthCacheAction::PasswordHashUpdate` to the code-paths that run after full token validation / successful authentication, so caching only occurs on `AuthResult::Success`. (added conditional `cache_action` and return of it on success paths in `src/common/src/idprovider/himmelblau.rs`).
- Hardened the resolver to reject any `AuthCacheAction::PasswordHashUpdate` that is paired with non-successful results, logging and returning an error for such illogical combinations. (updated `src/common/src/resolver.rs`).
- Changes are minimal and targeted to eliminate caching during `Next`/in-progress MFA and to preserve intended offline-breakglass caching after full success.

### Testing

- ✅ `cargo fmt --all -- --check` — formatting check passed after fixes.
- ✅ `git diff --check` — diff checks passed with no whitespace or trivial issues.
- ⚠️ `cargo test -p himmelblau_unix_common --lib` — attempted but blocked by missing system development files (`dbus-1` / `dbus-1.pc`) in the environment, so full test-suite could not complete.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a764ec8a28883269d5ec86315cae135)

### DIFF
--- a/src/common/src/idprovider/himmelblau.rs
+++ b/src/common/src/idprovider/himmelblau.rs
@@ -3435,13 +3435,6 @@ impl IdProvider for HimmelblauProvider {
                                 ));
                             }
                         };
-                        let action = match (
-                            self.config.lock().await.get_offline_breakglass_enabled(),
-                            password.as_ref(),
-                        ) {
-                            (true, Some(cred)) => AuthCacheAction::PasswordHashUpdate { cred: cred.clone() },
-                            _ => AuthCacheAction::None,
-                        };
                         *cred_handler = AuthCredHandler::MFA {
                             flow: Box::new($resp),
                             password,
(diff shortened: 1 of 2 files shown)


dmulder:
This is actually only an issue when we perform a DAG fallback after password entry has already completed.

Issue #1680, opened 2026-09-03
Learn short username mapping from successful full-UPN login


ghjklw:
### Summary

Learn and persist the short Unix username after a successful authentication using the full UPN, allowing subsequent logins with either form.

### Problem Statement

With `cn_name_mapping = true`, a short username normally needs to be mapped to a UPN before authentication.

This becomes problematic when a tenant has multiple UPN suffixes. On a fresh machine, "alice" alone may not provide enough information to determine whether the actual UPN is "alice@company.com" or another valid suffix.

The full UPN, however, is unambiguous. Once "alice@company.com" has successfully authenticated, Himmelblau has enough information to associate it with the Unix username "alice".

This is related to #179, but instead of requiring the mapping to be known before the first login, the mapping could be learned from a successful full-UPN login.

### Proposed Solution

With `cn_name_mapping = true`:

1. Allow the first login using the full UPN, e.g. "alice@company.com".
2. After successful authentication, derive the short username "alice".
3. Persist the relationship "alice@company.com ↔ alice" using the existing mapping cache.
4. On subsequent logins, resolve "alice" back to the learned UPN.
5. Continue allowing the full UPN at any time.

The mapping should only be learned after successful authentication.

Short names should not be assumed to be globally unique. If a proposed short name conflicts with another learned UPN or a local Unix account, Himmelblau should not create an ambiguous alias; the affected user can continue using the full UPN.

This seems like it could build on the existing `MappedNameCache`, `map_upn_to_name()` and `map_name_to_upn()` functionality.

If this approach fits the project direction, I would be happy to prepare a PR implementing it.

### Benefits

- Provides a clean first-login flow for tenants with multiple UPN suffixes.
- Avoids guessing a UPN suffix from a short username.
- Avoids external directory lookups from `name_mapping_script`.
- Keeps the full UPN as the canonical identity while exposing a conventional Unix username.
- Remains safe when short-name collisions exist by falling back to full-UPN login.

### Alternatives Considered

`name_mapping_script` can resolve a short username to a UPN, but on a fresh machine it still requires an external source of truth when the UPN cannot be inferred from the short name alone.

### Additional Context

Follow-up to #179, where UPN-to-local-name mapping and the need for reverse mapping through NSS were discussed.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct
---

### 💡 Help Make It Happen!

[...]

Issue #179, opened 2024-09-16
Ability to use a different object or change the UPN for username


tastle73:
This is probably an ER.

I would like to be able to use something other than the user@domain for the username.  Is this possible?

dmulder:
I think Entra ID requires the upn (user@domain) for authentication. I was considering mapping this to the cn though (so that you could leave off the @domain part). Is this what you're interested in?

tastle73:
That would work.  I have also seen SAML registered apps allow for the UPN to be munged from the AAD side.

dmulder:
This shouldn't be too difficult to add. It would get tricky FYI if you have multiple configured domains, because then we'd pick the first matching user. For example, if you have 2 domains `example.com` and `test.com` configured, then a user with the UPN `tux` in each domain (`tux@example.com` and `tux@test.com`), then we'd have to just pick the first user in the domain list.

tastle73:
Now that I think of it, that won't work for our use case. I need to be able to distinguish the AAD users from the local users and prevent username collisions.  What would probably be useful is a parser where I could use the cn + another string like ABCtom for example.

tastle73:
They will collide.  A lot of our usernames are not very unique

tastle73:
no, not at all.

dmulder:
Hrm, maybe I could introduce a filter option, which matches ABCtom then translates it to tom@abc.com, etc.

tastle73:
that would definitely work for us :)

dmulder:
I think I'm going to provide a script option, where you can provide a custom script that modifies the username as you wish.

tastle73:
let me know if you need someone to test

dmulder:
Ok, I'll probably have something for you to try today.

dmulder:
Hrm, I forgot that this could effect the user cache. This may take me a bit to sort out.

dmulder:
This is now partially implemented in #230 and #231. You now only need enter the CN name (the front part of the full UPN) to login.

dmulder:
@tastle73 I've added the script mapping feature to this branch: https://github.com/himmelblau-idm/himmelblau/tree/stable-0.6.x_map_logon_name

Would you try it out? I'll do some testing probably tomorrow.

tastle73:
this is the build exit:

dmulder:
@tastle73 there is no build output in your message. Did you forget to attach it?

tastle73:
[...]

PR #1685, opened 2026-09-04
Learn safe short UPN mappings after successful login

ghjklw:
## Summary

Follow-up to #179, solves #1680

The opt-in `learned_name_mapping` setting extends `cn_name_mapping` for users with alternate UPN suffixes. A user first signs in with the exact full UPN; later short-name login is available when that local part has been learned.

### Behavior

- Learning is active only when both `cn_name_mapping` and `learned_name_mapping` are enabled.
- Explicit full UPNs remain direct and valid.
- A successful exact full-UPN authentication learns its local part unless it is already claimed.
- The first learned UPN keeps the short alias. Later UPNs with the same local part remain full-UPN-only.
- Local Unix users are never learned as aliases.
- Learning is disabled for OIDC and when `name_mapping_script` is configured; script mappings retain their existing behavior.

### Implementation

Learned aliases use a separate persistent table from `name_mapping_script` cache entries. A unique index on `mapped_name` and `INSERT OR IGNORE` make first-writer alias ownership atomic. PAM retains the originally supplied username and records an alias only after `PAM_SUCCESS`.

### Validation

- `cargo fmt --all -- --check`
- `git diff --check`
- The targeted Rust test could not compile locally because the host lacks `libudev.pc`; upstream CI is the remaining validation boundary.

### DIFF
--- a/src/common/docs-xml/himmelblauconf/base/learned_name_mapping.xml
+++ b/src/common/docs-xml/himmelblauconf/base/learned_name_mapping.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<parameter name="learned_name_mapping"
+           section="global"
+           type="bool"
+           rust_type="bool"
+           documented="true"
+           domain_specific="false"
+           order="1710">
+<description>
+Enables short Unix names that Himmelblau learns only after a successful login with the exact full UPN. When enabled, a short name is unavailable until that first full-UPN login succeeds. It replaces the implicit primary-domain fallback from cn_name_mapping, preventing an unobserved primary-domain user from colliding with a learned alias. The first UPN learned for a local part keeps its short name; later UPNs with the same local part continue to use their full UPN. This option is ignored when name_mapping_script is configured.
+</description>
+<conf_description>Learn short names after successful full-UPN login; require a full UPN for first login.</conf_description>
+<default>false</default>
+<example>learned_name_mapping = true</example>
+</parameter>
--- a/src/common/src/config.rs
+++ b/src/common/src/config.rs
@@ -853,35 +853,54 @@ impl HimmelblauConfig {
         None
     }
 
+    fn is_local_unix_name(account_id: &str, passwd_path: &str) -> bool {
+        let mut contents = vec![];
+        if let Ok(mut file) = File::open(passwd_path) {
+            let _ = file.read_to_end(&mut contents);
+        }
+        parse_etc_passwd(contents.as_slice())
+            .unwrap_or_default()
+            .into_iter()
+            .any(|u| u.name.to_string() == account_id)
+    }
+
     /// Convert a login name to a UPN, or `None` if it cannot be a directory user
-    /// (empty, or a known local-only name). Callers treat `None` as user-unknown
-    /// and must not look it up against Entra. See himmelblau-idm/himmelblau#1392.
+    /// (empty, a known local-only name, or an ambiguous learned mapped name).
+    /// Callers treat `None` as user-unknown and must not look it up against Entra.
+    /// See himmelblau-idm/himmelblau#1392.
     pub fn map_name_to_upn(&self, account_id: &str) -> Option<String> {
         self.map_name_to_upn_impl(account_id, "/etc/passwd")
     }
 
-    /// Inner impl with an injectable passwd path for hermetic unit tests.
+    /// Inner impl with an injectable passwd path for existing hermetic unit tests.
     fn map_name_to_upn_impl(&self, account_id: &str, passwd_path: &str) -> Option<String> {
+        let name_cache = if passwd_path == "/etc/passwd"
+            && (self.get_name_mapping_script().is_some() || self.get_learned_name_mapping())
+        {
+            MappedNameCache::new(MAPPED_NAME_CACHE, &Mode::ReadWrite).ok()
+        } else {
+            None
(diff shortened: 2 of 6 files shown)


aidangarske:
Nice work, this is close. A few things need sorting before merge

src/common/src/config.rs
```diff
@@ -936,43 +952,133 @@ impl HimmelblauConfig {
                         );
                     }
                 }
+            } else if learned_name_mapping {
```
aidangarske, line 955: Both resolution paths at config.rs:955 use raw `learned_name_mapping` instead of insertion’s effective predicate, breaking CN-off, OIDC, and failing-script fallbacks. Can you centralize that predicate for cache opening, learned lookups, and fallback suppression, and add regression tests for those cases?

src/pam/src/pam/mod.rs
```diff
@@ -475,13 +479,19 @@ impl PamHooks for PamKanidm {
...
+            cfg.clone(),
             &account_id,
             &service,
             opts,
             msg_printer,
         );
 
+        // Preserve the original PAM username and only learn after a successful
+        // authentication of an explicitly supplied full UPN.
+        if should_learn_short_name_after_auth(&result, &supplied_account_id) {
```
aidangarske, line 491: src/pam/src/pam/mod.rs:491 writes from the PAM caller, so non-root `getuid()` leaves the cache read-only while `insert_learned_mapping` returns success without persisting the alias. Can you move the write to the privileged tasks service, surface skipped or failed writes, and cover successful auth from a non-root PAM caller?

src/common/src/config.rs
```diff
@@ -936,43 +952,133 @@ impl HimmelblauConfig {
...
+            || !supplied_name.eq_ignore_ascii_case(authenticated_upn)
+        {
+            return Ok(());
+        }
+
+        let authenticated_upn = authenticated_upn.to_lowercase();
+        let Some((short_name, _)) = split_username(&authenticated_upn) else {
+            return Ok(());
+        };
+        if short_name.is_empty() || Self::is_local_unix_name(short_name, passwd_path) {
```
aidangarske, line 1026: Alias learning at config.rs:1026 checks /etc/passwd but not NSS_IGNORE_EXACT or NSS_IGNORE_PREFIX, so it can learn aliases that map_name_to_upn() deliberately rejects and break round trips. Can you apply the same local-only-name policy when learning aliases and add exact and prefix-reserved tests?

src/common/src/config.rs
```diff
@@ -936,43 +952,133 @@ impl HimmelblauConfig {
...
+            let cn_name_mapping = self.get_cn_name_mapping();
+            let learned_name_mapping = self.get_learned_name_mapping();
+
+            if name_mapping_script.is_some() {
+                return name_cache
+                    .map(|cache| cache.get_mapped_name(upn))
+                    .unwrap_or_else(|| upn.to_string());
+            }
+
+            if learned_name_mapping {
```
aidangarske, line 1061: config.rs:1061 exposes cached aliases without rechecking the current local namespace, so a later local account can own the returned name. Can you recheck before returning one, use the full UPN after an ownership change, consider removing the stale row, and add the lifecycle test?

src/common/src/config.rs
```diff
@@ -853,35 +853,54 @@ impl HimmelblauConfig {
...
+    /// Callers treat `None` as user-unknown and must not look it up against Entra.
+    /// See himmelblau-idm/himmelblau#1392.
     pub fn map_name_to_upn(&self, account_id: &str) -> Option<String> {
         self.map_name_to_upn_impl(account_id, "/etc/passwd")
     }
 
-    /// Inner impl with an injectable passwd path for hermetic unit tests.
+    /// Inner impl with an injectable passwd path for existing hermetic unit tests.
     fn map_name_to_upn_impl(&self, account_id: &str, passwd_path: &str) -> Option<String> {
+        let name_cache = if passwd_path == "/etc/passwd"
```
aidangarske, line 877: config.rs:877 opens the SQLite cache and runs schema checks before excluding empty, local, ignored, or full-UPN names, adding DB work to routine NSS/PAM lookups. Can you construct it lazily only in the script-insertion or learned-short-lookup branch, and keep schema migration separate from routine lookups?

src/common/docs-xml/himmelblauconf/base/learned_name_mapping.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<parameter name="learned_name_mapping"
+           section="global"
+           type="bool"
+           rust_type="bool"
+           documented="true"
+           domain_specific="false"
+           order="1710">
+<description>
+Enables short Unix names that Himmelblau learns only after a successful login with the exact full UPN. When enabled, a short name is unavailable until that first full-UPN login succeeds. It replaces the implicit primary-domain fallback from cn_name_mapping, preventing an unobserved primary-domain user from colliding with a learned alias. The first UPN learned for a local part keeps its short name; later UPNs with the same local part continue to use their full UPN. This option is ignored when name_mapping_script is configured.
```
aidangarske, line 10: learned_name_mapping.xml:10 only documents the `name_mapping_script` conflict, but activation also requires `cn_name_mapping` and no generic OIDC configuration. Can you document all three constraints in the XML and regenerate the man page and Nix option docs?

